PR #313, opened 2013-03-14
HTML comments with either an element callback (EJS) or a helper (Mustache) rendered incorrectly

scorphus:
In a Mustache template, for example, if you have:

```
<li {{data "item"}}>
    <!-- this is comment #1 -->
    {{title}}
    <!-- this is comment #2 -->
</li>
```

this gets incorrectly rendered:

``` html
<li  data-view-id='9'>
    <!-- this is comment #1 - data-view-id='10'/>
    One
    <!-- this is comment #2 -->
</li>
```

This bug is demonstrated, for both EJS and Mustache, in this fiddle: http://jsfiddle.net/scorphus/vmLvA/

The fix stops the effect of an element callback (in EJS templates) or a helper (in Mustache templates) and is demonstrated in this fiddle: http://jsfiddle.net/scorphus/vmLvA/1/

### DIFF
--- a/view/ejs/test/ejs_test.js
+++ b/view/ejs/test/ejs_test.js
@@ -1283,4 +1283,33 @@ test('live binding with html comment', function(){
 	equals(div.getElementsByTagName('table')[0].getElementsByTagName('td').length, 0, '0 items in list');
 })
 
+test("HTML comment with element callback", function(){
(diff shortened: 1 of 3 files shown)


amirJahani1993:
[](url)

Issue #344, opened 2013-04-03
Adding .get() and .remove() to Model.List, putting .slice(), .grep(), .match() in observe/list


roelmonnens:
.get(), .remove(), .slice(), .grep(), .match() are now in models/list/list.js.

It seems more correct that .get() and .remove() should be in the original Model.List and not in model/list/list.js (this extends Model.List).

model/list/list.js also contains .slice(), .grep() and .match(). This would be better in a the plugin observe/list.
observe/list already exists and contains the .filter() and .map() functions.

I add this issue on request of daffl with whom I have discussed this on irc.

PR #345, opened 2013-04-04
adding .slice() .grep() and .match() to observe/list; adding .get() to can.Model.List

roelmonnens:
These methods are available in model/list but can be moved to
observe/list.

linked to issue #344

.remove is not implemented because you can use .get() to get the model instance and then call .destroy()

.grep() and .match() are removed because .filter() is the same.


julia-allyce:
This PR hasn't been worked on in a long time. I am closing it. Please feel free to reopen or update it if we want to reconsider it.

daffl:
We were thinking of adding Lodash/Underscore methods to can.List in CanJS 3.

PR #584, opened 2013-12-01
Update pushstate.js ( Fix for firefox push/replacestate override problem...

Helidium:
...)

Hi I was using canjs with new pushstate functionality. All is working well in chrome / opera but not in firefox. The browser stopped listening to url changes all of the sudden after a couple of seconds of inactivity. The most bizzare situation which I am still not familiar with was when I changed any map property with attr the setState was called, otherwise not.

Anyway I deleted the push/replacestate override and call the setState in the click handler.


pYr0x:
no test, 2 year past, close this.
feel free to reopen if needed

PR #740, opened 2014-02-16
Improve can.scope and $.fn.scope code and add assignment support

zkat:
This patch makes it so all of the following are supported:

``` js
$(el).scope() // => can.Map instance
$(el).scope("foo") // => fetches "foo" property of can.Map instance
$(el).scope("foo", value) // => sets value as the value for the "foo" property on that instance
```

as well as the `can.scope()` equivalents.

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -433,26 +433,62 @@ steal("can/util", "can/view/callbacks","can/control", "can/observe", "can/view/m
 		}
 	});
 
-	// If there is a `$` object and it has the `fn` object, create the `scope` plugin that returns
-	// the scope object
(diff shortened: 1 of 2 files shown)


zkat:
@daffl all done

component/component.js
```diff
@@ -418,26 +418,59 @@ steal("can/util", "can/view/callbacks","can/control", "can/observe", "can/view/m
...
+			return scope.attr(attr);
+		default:
+			scope.attr(attr, val);
+			return el;
+		}
+	};
+
+	// If there is a `$` object and it has the `fn` object, create the
+	// `scope` plugin that returns the scope object.
+	if (window.$ && $.fn) {
```
daffl: I think this needs to be `window.jQuery`

PR #761, opened 2014-03-04
Event "Bubbling" for can.Model subclasses

bmomberger-reciprocity:
I've been perpetually annoyed at defining something like the following every time I listen on a Model class event:
```javascript
"{TodoTask} updated" : function(model, ev, instance) {
  if(instance instanceof TodoTask) {
 // ... actually get to work
 }
}
```

This is because of a peculiarity of how can.Construct creates and propagates static properties.  The __bindEvents repository for event bindings on objects gets passed along if it gets defined before the subclass is created.

That is to say, if I create ```can.Model("Todo", { init : function() { this.bind("created", ...); } }, {})``` and then extend Todo into SharedTodo, both classes share the same created/updated/destroyed event bindings.  All subclasses of Todo will share these event bindings and you have to pick apart which type you actually want to listen on, on each event firing.

This pull request is intended to resolve this problem elegantly.  Now event bindings can propagate up a class hierarchy instead of spanning the entire tree.

### DIFF
--- a/model/model.js
+++ b/model/model.js
@@ -948,6 +948,10 @@ steal('can/util', 'can/map', 'can/list', function (can) {
 				// Add ajax converters.
 				can.Model._reqs = 0;
 				this._url = this._shortName + '/{' + this.id + '}';
+				this.__bindEvents = {};
+				if(base.__bindEvents) {
+					this.__bindEvents.__base = base.__bindEvents;
+				}
 			},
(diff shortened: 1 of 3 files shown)


julia-allyce:
@bmomberger-reciprocity @thomblake I am closing this issue its been over a year since its last activity. Please reopen or update if there is still interest in making this contribution.

Issue #773, opened 2014-03-11
can.Object.same: compare logic is incorrect when it comes to dates


dispatchrabbi:
When comparing objects with properties that have dates in them, same() is incorrect and inconsistent:

``` js
var a = {foo: null};
var b = {foo: new Date()};
var c = {foo: new Date(b.foo.getTime())};

can.Object.same(a, b); // should be false; is true
can.Object.same(b, a); // should be false; is false (but not the same as same(a, b)!)
can.Object.same(b, c); // should be true, is false
```

Here's a demonstrative fiddle: http://jsfiddle.net/bu9sd/

PR #788, opened 2014-03-14
can.Object.same should check date values when comparing

ccummings:
This is related to #773 
When values in an Object are dates, their time values should be compared since `new Date() === new Date() //-> false`

http://jsfiddle.net/D9kEv/

This involves a breaking API change.


justinbmeyer:
this change will need to be made here: https://github.com/canjs/can-set/issues/35

PR #888, opened 2014-04-16
Ensure can-value bound input stay in sync with compute

### LINKED ISSUE #887: can-value becomes out of sync when a compute rejects the new value
http://jsfiddle.net/6PeBb/1/

This demo includes a compute that rejects non-numbers. If you type in a non-number value the compute's setter will return the old value.

This means the `<input>` and the compute are not in sync.

matthewp:
It's possible for a `can-value` bound input to become out of sync with it's underlying data if that data is a compute with a setter function. The setter function might reject the user's input or set the value to something else entirely.

Previously we were assuming that the compute's new value would always be the same as the input's value. This checks to make sure that the values are the same, otherwise it sets the input's value to reflect the *actual* value of the compute.

Fixes #887

### DIFF
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -236,8 +236,15 @@ steal("can/util", "can/view/mustache", "can/control", function (can) {
 			if (!this.element) {
 				return;
 			}
+			var el = this.element[0];
+			
 			// Set the value of the attribute passed in to reflect what the user typed
-			this.options.value(this.element[0].value);
+			var newVal = this.options.value(el.value);
+			
+			// If the newVal isn't the same as the input, set it's value
+			if(el.value !== newVal) {
+				el.value = newVal;
+			}
 		}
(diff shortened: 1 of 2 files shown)


view/bindings/bindings.js
```diff
@@ -236,8 +236,15 @@ steal("can/util", "can/view/mustache", "can/control", function (can) {
 			if (!this.element) {
 				return;
 			}
+			var el = this.element[0];
+			
 			// Set the value of the attribute passed in to reflect what the user typed
-			this.options.value(this.element[0].value);
+			var newVal = this.options.value(el.value);
```
justinbmeyer: I don't think can.compute necessarily always returns its current value when being set.  I would change this to:

```
this.options.value(el.value)
var newVal = this.options.value();
```

daffl:
Resubmitted as #1436

PR #918, opened 2014-04-29
Providing more control over how can.view resolves templates

### LINKED ISSUE #314: Make can.view not make an AJAX request when an id lookup fails
There should be a way to tell can.view to stop if it doesn't find an element with the given id. 

Right now can.view will try first to lookup a template by id and fallback to making an AJAX request. https://github.com/bitovi/canjs/blob/master/view/view.js#L471

I have two ideas for solving this:

can.view takes an Object with an id or url key (it can also take a String which will do an id lookup then Ajax)

```js
can.view({id: 'tmpl'}) //-> id lookup, no Ajax request
can.view({url: 'mytmpl/tmpl.mustache'}) //-> Ajax request, no id lookup
can.view('tmpl') //-> id lookup then Ajax request
```

Another way to know is to assume a `#` at the beginning of the URL means id lookup only. We already assume as much in the [code](https://github.com/bitovi/canjs/blob/master/view/view.js#L486), but still fallback to an Ajax request. 

```js
can.view('#tmpl', {}) //-> id lookup only
can.view('tmpl', {}) //-> id lookup then Ajax request
```

ccummings:
Proposal to resolve #314 

There is one breaking API change here which is not performing an AJAX call when a `#` is in the provided URL. But the rest of the fix would be new API that is backwards compatible. 

I added `can.view.forceLocalLookup` to make it easier for existing codebases to change over to local lookups. Some apps use URLs or script tags in different environments and this lets them toggle between the two easily.

To get this into a minor release (before 3.0) we could remove the breaking change and add it later.

### DIFF
--- a/view/doc/view.md
+++ b/view/doc/view.md
@@ -21,13 +21,38 @@ a [https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment documentFra
                 return first +" "+ last
             }
         });
-        
+
     document.getElementById('contacts').appendChild(frag)
 
-@param {String|Object} idOrUrl The URL of a template or the id of a template embedded in a script tag or an object containing a `url` property for the URL to load and an `engine` property for the view engine (`mustache` or `ejs`) if it can't be infered from the file extensions or script tag type.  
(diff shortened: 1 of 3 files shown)


justinbmeyer:
Closing as `stache.from` does not do ajax requests. This is gone in 3.0.

Issue #951, opened 2014-05-03
can.Model promises


retro:
We should add promises to can.Model similar to can.List. API could be something like:

    Model.get({ ... })

This would internally call `findOne` and it should return an empty model which will be filled with the data from the API, but I think that this clashes with how the model store works, and it could break bindings.

justinbmeyer:
closing.  once we added promise support in our templates, the importance of this faded a bit.

PR #956, opened 2014-05-03
951 can model promises

retro:
This implements promise API for can.Model as mentioned in #951 . Docs are missing, but I'll add them if we agree on the implementation.

It adds the `.get` static function on the model which calls `.findOne` internally and returns a compute that holds the deferred object returned from the `.findOne` function. When this deferred is resolved or rejected compute is updated.

It also implements deferred state for the model instances which gets updated when `destroy` or `save` request is happening.

It has a demo similar to one in the `can/list/promise` plugin.

### DIFF
--- a/model/promise/promise.html
+++ b/model/promise/promise.html
@@ -29,7 +29,7 @@
   {{/if}}
 
   {{#if user.isRejected}}
-    We were unable to load the user because {{user.reason.responseJSON.message}}
+    {{user.reason.responseJSON.message}}
   {{/if}}
 
 </site-app> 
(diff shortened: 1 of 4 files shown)


justinbmeyer:
@retro is this needed now that we have observable promises?  I'm closing, please let me know if we need to reopen.

pYr0x:
i do not found any documentation of that?

justinbmeyer:
I'm not sure where we would document this. Possibly stache.keys. The promises themselves are not observable by can.compute. Instead, the template engine creates a shim of the promise that is observable.

daffl:
I was thinking of just creating a "Promises" page (in the Pages section at http://canjs.com/docs/can.stache.html) that talks about how to use it.

PR #978, opened 2014-05-08
Added breaking test of string attributes not working with scope retured ...

whitecolor:
...as function (for mustache engine)

in scope "string" attribute defined as '@' and should be equal to passed value in template which is "string".

But it equals "@".

It seems to work with stache though.


pYr0x:
i think we can close this.

Issue #179, opened 2012-11-28
Make promises observable


justinbmeyer:
[VOTE HERE](http://bithub.com/event/25912)
```
{{#Todo.findAll}}
  <li>
    {{name}}
  </li>
{{/}}
```

matthewp:
It will be good to have a way to know when a template is "done", after all promises have been resolved.

justinbmeyer:
You can always wire that up yourself with `can-inserted` like:

```
{{#if promise.isRejected}}<span class='rejected' can-inserted="{doSomething}">{{promise.reason.message}}</span>{{/if}}
```

matthewp:
I'm thinking of the server rendering use-case, would want something more generic.

daffl:
I think we can close this, Observable promises made it into 2.2.

PR #1031, opened 2014-05-30
Mustache deferreds

retro:
Implements `def` helper that can handle deferreds in the template code. 

Sample use:

	{{#def Todo.findAll}}
		{{#if isResolved}}
			{{@data}}
				... iterate over the data
			{{/}}
		{{/if}}

		{{#if isRejected}}
			{{@error}}
				... print out the error
			{{/}}
		{{/if}}

		{{#if isPending}}
			still pending
		{{/if}}
	{{/def}}

I prefer this syntax to anything other mentioned in the #179 because it's more explicit and it's immediately clear what's happening.

`def` helper can also handle non-deferred values in which case it will treat it as resolved.

@justinbmeyer any comments?

### DIFF
--- a/view/mustache/mustache.js
+++ b/view/mustache/mustache.js
@@ -2056,6 +2056,31 @@ steal('can/util',
 						console.log(expr, options.context);
 					}
 				}
+			},
+			'def' : function(deferred, opts){
+				var resolved = can.compute(),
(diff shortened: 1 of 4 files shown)


daffl:
This has been implemented in 2.2, closing this PR.

PR #1051, opened 2014-06-04
less noisy dev warnings

moschel:
1) only show the construct.extend warning once per page
2) only show each mustache key missing warning once per key (if this is in a loop it gets very noisy)
3) only show mustache key warnings for rendering properties, not for arguments to a helper
4) mustache key warnings were broken and showing up for all helper access also


andrejewski:
Closing as retro 👴

PR #1114, opened 2014-06-25
Sorting a List doesn't update in a template correctly

asavoy:
I'm trying to use `.comparator` and `.sort()` as described in [can.Map.prototype.sort](http://canjs.com/docs/can.List.prototype.sort.html) to change the sorting of a can.List that is bound in a Mustache template.

The can.List gets correctly sorted, but there appear to be a few bugs that prevent the template from reflecting the sorted data correctly. 

Please refer to this [JSFiddle](http://jsfiddle.net/qYdwR/1627/):

* Calling `list.sort()` won't update the rendered template at all.
* But calling `can.trigger(list, 'length')` or `list.push(newItem)` WILL force the update, but only for `{{#key}}...{{/key}}` syntax. It's not the worst workaround, but I prefer to use `{{#each}}` because that provides `{{@index}}`.
* The `{{#each}}` syntax will never show list sorted, even after adding items to it.

  * (Another weird thing is, in this case, adding an item will render the new item twice.)

I've added tests that cover all of these issues.

### DIFF
--- a/list/list.js
+++ b/list/list.js
@@ -718,7 +718,7 @@ steal("can/util", "can/map", "can/map/bubble.js",function (can, Map, bubble) {
 				// Call the original method.
 				res = orig.apply(this, args);
 
-				if (!this.comparator || args.length) {
+				if (!this.comparator && args.length) {
 
 					this._triggerChange("" + len, "add", args, undefined);
 				}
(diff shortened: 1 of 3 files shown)


daffl:
This will be fixed via #1454 in 2.2.

PR #1121, opened 2014-06-26
fix #1108

### LINKED ISSUE #1108: can.stache dosent render custom tags with Colon  ":"
can.stache destroy the custom tag with colon (eg: x:widget) it treats "x" as the custom tag and ":widget" as an attribute which is not the case with can.mustache here is a fiddle about the case http://jsfiddle.net/cherif_b/YGbrh/

cherifGsoul:
#1108

### DIFF
--- a/view/parser/parser.js
+++ b/view/parser/parser.js
@@ -1,266 +1,267 @@
 /* jshint maxdepth:7*/
-steal("can/view", function(can){
-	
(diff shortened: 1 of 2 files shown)


daffl:
Closing this as a duplicate of #1287 which already made it into 2.1.4.

PR #1158, opened 2014-07-09
Make can/util/fixture work with new Steal

matthewp:
This makes can/util/fixture work with the new Steal by replacing a call to the join Steal URI `mapJoin` with the new  `steal.joinURIs`.


daffl:
@matthewp is this relevant for latest minor? If so can you fix and resubmit?

daffl:
This is fixed in latest minor.

PR #1186, opened 2014-07-31
Added documentation for the can/map/list plugin

### LINKED ISSUE #1165: Incomplete documentation for can.List.prototype.filter
It probably bears mentioning in the docs that the prototype method `filter` of `can.Map.List` overrides that of `can.List`, and does not take a second argument `thisArg` as the method in `can.List` does.

andykant:
The `can/map/list` plugin provides live-updating filtered and mapped lists.

These changes add documentation for that plugin, as well as a note that it overwrites the original `can.List.prototype.filter` with a new method signature. Fixes #1165


daffl:
Closing and resubmitting against minor.

PR #1199, opened 2014-08-13
Pass events for can-event

### LINKED ISSUE #1195: can-event doesn't pass extra arguments
Normally when triggering an event with jQuery I can do:

	$(element).trigger('selected', val);
	
`val` is received by the listener as an extra argument. However with can-event this doesn't happen, extra arguments are discarded.

This is useful when I want to trigger my own events, please see this fiddle;

http://jsbin.com/koxobamo/13/edit

asavoy:
Fixes #1195


daffl:
I agree. This doesn't break any other libraries where passing event arguments isn't possible anyway and it fixes an actual issue. Will go into 2.2, thanks!

PR #1226, opened 2014-09-09
can.compute placed on a can.Map should not be readonly

john-charles:
This is a failing test to go with the bug report by the same name.

When a compute is added to a base can.Map.extend the compute becomes read only.


daffl:
This can be solved with the define plugin by setting `{ type: 'compute' }`

Issue #1265, opened 2014-10-09
can.List.prototype.sort is not working


stevenvachon:
https://github.com/bitovi/canjs/pull/1114 seems to be almost there (?)

Although, that appears to sort a clone and reset the value. Shouldn't the current plugin work, only trigger the correct event? What would that event be?

daffl:
This has been fixed via #1454

PR #1267, opened 2014-10-09
Added simple can.List.prototype.sort to core (different from plugin)

stevenvachon:
This does not close #1114 or #1265, which I'd originally set out to do. Instead of fixing the sort plugin, I decided to write a simple sort in core. Reasons:

* It's not very often that we need auto-sort
* It's not very often that we need `"move"` events

Both of these add complexity and performance overhead that belong in the optional plugin.

This PR:

* Uses two total array/Map/List operations: clone<sup>1</sup> and sort ... no set
* Triggers `"remove"` and `"add"` events<sup>2</sup> for DOM mutations
* Triggers `"sort"` event *first* for user listeners to distinguish between `"add"`/`"remove"`
* Renames the old sort plugin to "autosort", as to prevent errors in tests and to avoid confusion
* Gives us a fast and simple sort, good for most things
* Offers these syntaxes:
  * `sort()`
  * `sort( function(a,b){ ... } )`
  * `sort("key")`, `sort("deep.key")`
  * `sort({ comparators:["deep.key","deep.other"], order:"descending" })`

### Notes
1. I figured out a way to avoid cloning, but it resulted in the `"sort"` event accessing outdated data (see [comment](https://github.com/bitovi/canjs/pull/1267#issuecomment-58798792)).
2. If these events could be triggered silently, `"sort"` could then be the only public event.


daffl:
I think we can close this one since we have the new sort plugin in 2.2.

PR #1280, opened 2014-10-20
Rename _setupDefaults to _getDefaults and use it in LazyMap

w3dot0:


daffl:
There unfortunately seem to be a bunch of conflicts with some of the changes we recently made. Good catch with using `_setupDefaults` though. I submitted a separate PR (#1457) with using those (I'm kind of hesitant renaming it even if it is technically a "private" method).

PR #1357, opened 2014-12-05
Argument support for can-EVENT

### LINKED ISSUE #1219: can-EVENT arguments
[VOTE HERE](http://bithub.com/event/44023)

It should be possible to pass `can-EVENT` bindings arguments from the scope like:

```html
<li can-click="{doSomething key1 ../key2 "foo" @element @event}">
```

Pull request from @zkat : #1357

zkat:
can-EVENT now parses its arguments similar to how stache helpers handle it and prepends those arguments and the hash to the arg list.

See #1361 and closes #1219


zkat:
@daffl added tests, docs.

@justinbmeyer added and documented `@element` and `@event`.

view/bindings/bindings_test.js
```diff
@@ -24,24 +25,49 @@ steal("can/view/bindings", "can/map", "can/test", "can/view/stache", function (s
...
 		can.trigger(p0, "click");
 
+
+		template = can.view.mustache("<div>" +
+			"{{#each foodTypes}}" +
+			"<p class='with-args' can-click='{withArgs @event @element title content=content}'>" +
+			"{{content}}</p>" +
+			"{{/each}}" +
+			"</div>");
+		function withArgs(ev1, el1, title, hash, foodType, el2, ev2) {
```
justinbmeyer: Once a user supplies arguments, should the "old defaults" be added?  Should `foodType`, `el2`, `ev2` be passed?

I'm thinking of people checking arguments.length sorta thing.

justinbmeyer:
@zkat Amazing work on this pull request.  From the code, I'm not sure if the current component's values will be available as arguments or not.  For example:

```js
can.Component.extend({
  tag: "can-info",
  scope: {
    data: {message: "HI"}
  }
})

var tmpl = can.stache("<can-info can-click="{printOut data.message}"></can-info>")
tmpl({
  printOut: function(text){
    console.log(text);
  }
});
```

There's certainly a need for a component to be able to expose its own data to event handlers like this.  But as you're an advocate of lexical semantics, I can see you being a bit uneasy with this.

I think we need a way to enable this "now" and have a good plan on enabling this in a better way later.

view/bindings/bindings_test.js
zkat: I like this change a lot actually. Specifically, it means we don't need to do anything special for something like `{@root.attr name=foo email=bar}` to work, whereas if we continue passing in the extra arguments when explicit arguments are present, we would end up calling the equivalent of `viewmodel.attr(valuesHash, true)` and replace everything, which is not very friendly :(

Shall I make this change? (or can someone else? I'm on PTO today)

justinbmeyer: I'm on PTO too and Monday I'm doing a new hire training.  So either @daffl or you maybe.

zkat:
@daffl I've made all the changes, but the amd tests specifically are failing and I don't understand why. Can you take a look? The regular tests pass just fine. :\

daffl:
I re-submitted this against master in #1495. Apparently I fail at rebasing to I'll just merge into master and then update minor with it.

PR #1367, opened 2014-12-11
Get travis to run test suites in parallel

daffl:
Implements parallel test runs by spinning off a Travis VM for each build which improves the CI runtime from ~20 minutes to ~5.


daffl:
This didn't turn out to be as effective as expected. On average waiting for several Travis VMs takes more time than running everything in series. Maybe different with a paid account.

Issue #1037, opened 2014-06-02
dependencies in templates


justinbmeyer:
I'd like templates to be able to declare dependencies.  For example, a template depends on a component.

PR #1369, opened 2014-12-15
can.stache import support

zkat:
This adds support for a variety of import syntaxes when using `steal` to load `stache` templates. See #1037 

The acceptable syntaxes are:

* `{{@import "modulename"}}`
* `<link rel="import" href="modulename">`

Additionally, this adds `steal`-based static resolution for partials, so the following is also supported, and will inline the template:

* `{{>modulename.stache!}}`

### Notes and Justification
The implementation is fairly rough right now, using plain old regexes to parse out and remove import information from the module string when the file is parsed by stache. Ideally, this would use our stache parser to yank these out more reliably, specially the `<link>` tags.

The biggest use case for this patch, aside from making partials more useful in applications that use `steal` as a loader, is to import `can.Component` and helper definitions right in the `stache` templates that use them without having to worry about load order of modules (and having these in the file that actually uses them).

### Example
```html
<link rel="import" href="can-tabs">
<link rel="import" href="can-tabs-tab">
{{@import "my_stache_helpers}}
<can-tabs>
  {{#each tabDefs}}
  <can-tab title="{{capitalize name}}">
    {{>./tab-boilerplate.stache!}}
    {{content}}
  </can-tab>
  {{/each}}
</can-tabs>
```

### DIFF
--- a/view/stache/system.js
+++ b/view/stache/system.js
@@ -1,4 +1,4 @@
-steal("can/view/stache", function(stache){
+steal("can/view/stache", "can/view/scope", function(stache, Scope){
 
 	var escMap = {
 		'\n': "\\n",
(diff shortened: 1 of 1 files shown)


zkat:
See #1386

PR #1371, opened 2014-12-16
booleanAttr for components

zkat:
Components that want to use HTML-style booleans are currently forced to write this function manually. I think this is valuable enough to can.Component users that it's good to have built-in to the define plugin.


daffl:
Looks good. Do we need to add a test for this though?

justinbmeyer:
And docs.

zkat:
@daffl done

PR #1372, opened 2014-12-16
updating module names for builder

justinbmeyer:
updates module names for new builder


daffl:
Looks like the history on this one is borked but this is included in #1440 via https://github.com/bitovi/canjs/commit/c820813e07c017a6c7de5b4d874a3a128e7e4d33 now anyway.

PR #1380, opened 2014-12-23
Namespacing Fix

DVSoftware:
We use namespacing a lot accross the application, and we encountered an issue when we create for example:
Model.Leave.Type, and then Model.Leave
This will cause the initial Model.Leave object containing the Type to be overwritten by Model.Leave constructor.
The fix will extend the new constructor with initial object properties.

### DIFF
--- a/construct/construct.js
+++ b/construct/construct.js
@@ -1,9 +1,9 @@
 // steal-clean
 steal('can/util/string', function (can) {
 	// ## construct.js
-	// `can.Construct`  
+	// `can.Construct`
 	// _This is a modified version of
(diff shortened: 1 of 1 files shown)


daffl:
This looks like it makes sense and can go into 2.2. Is there a chance you can add some tests?

DVSoftware:
Sure, i'll add them in a few days
18.02.2015. 21.31, "David Luecke" <notifications@github.com> је написао/ла:

julia-allyce:
Hey @DVSoftware are you still interested in adding tests and making this contribution?

Issue #1390, opened 2015-01-05
can-autorender


justinbmeyer:
Based on: https://github.com/bitovi-components/component-utils/blob/master/autoloader.js, I'm proposing a `can-autorender` attribute that will be used to automatically render templates.  For example:

```html
<body>
  <script type='text/stache' can-autorender id='demo'>
    <h1>{{message}}</h1>
  </script>
  <script src='canjs.js'></script>
  <script>
    $("#demo").scope().attr("message", "Hello World");
  </script>
</body>
```

Would result in:

```html
<body>
  <script type='text/stache' can-autorender id='demo'>
    <h1>{{message}}</h1>
  </script>
  <h1>Hello World</h1>
  <script src='canjs.js'></script>
  <script>
    $("#demo").scope().attr("message", "Hello World");
  </script>
</body>
```

This should be available on normal elements.  Consider:

```html
<body can-autorender type='text/stache'>
  <h1>{{message}}</h1>
  <script src='can.js'></script>
  <script>
    $("body").scope().attr("message", "Hello World");
  </script>
</body>
```

If `can-autorender` is on a non-script element, any scripts within that element should NOT be rendered as part of the template.

## Rendering

The embedded templates should be rendered on dom content loaded.

If the source element is in the `<head>` the rendered result will be appended to the `<body>`.

## Scope

If `can-autoload` is placed on a script tag, a `can.Map` instance will be available on the script element:

```js
$("script#demo").scope() //-> can.Map
```

It's possible that the scope will want to be accessed prior to rendering.  For this reason, can.scope should be made to create a can.Map instance if one does not already exist.  

## Other Considerations

- We might need to patch DOMContentLoaded for other libraries.
- We might want to specify a src so these templates can be loaded outside the page.

cc @sykopomp

justinbmeyer:
I like `can-render` a bit more than autoload.  However, the plugin might be `can/view/render` which would conflict with the `can.view.render` function.  

We could put it under bindings as `can/view/bindings/render`.

justinbmeyer:
Here's what the slider might look like with autoload:

```html
<script type='text/stache' can-autoload>
    <can-import from="can-slider"/>
	<can-slider></can-slider>
</script>
<script src="./node_modules/steal/steal.js" main="can/view/autoload"></script>
```

with render:

```html
<script type='text/stache' can-render>
    <can-import from="can-slider"/>
	<can-slider></can-slider>
</script>
[...]

Issue #1387, opened 2015-01-02
can.stache will import components it depends on


justinbmeyer:
Related to #1386, `can.stache` templates should be able to import components and modules they depend on.

For example, if a stache template looks like:

```
<div><can-slider></can-slider></div>
```

It will be transpiled to import `can-slider`.

Also, a template might depend on helpers.  These can be imported like:

```
<import from="my/helpers">
```

It's possible that a component's name doesn't match it's moduleName.  We can also support the following:

```
<import from="components/slider" as="can-slider">
<can-slider></can-slider>
```

In this case `can-slider` will not be imported, only `components/slider`.

### Other

 - This will only happen for templates loaded with the can/view/stache/system extension.
 - It's possible that we eventually support this happening with progressively loaded templates.  However, can.view will need to support promises.

Issue #1386, opened 2015-01-02
can.import


justinbmeyer:
Upcoming features of CanJS will need integration for importing modules.  Examples are `can-app`, and `can.stache` having the ability to import components it depends on.

I'm proposing a `can.import` which will work very similar to `System.import`.  It will take a moduleName and return a promise that is resolved with the moduleName:

```
can.import("can-slider").then(function(CanSlider){
  
})
```

The default behavior of `can.import` will be to use the `System` loader, if that doesn't exist, it will use AMD, and then CJS.

Other module loaders (such as StealJS) might overwrite `can.import` to make sure the config has finished loading.

whitecolor:
Is it possible to make it load multiple modules ```can.import("can-slider", "some-other")``` ?

matthewp:
What will be the use-case for this?

whitecolor:
Well what is the use case for this:

```
can.import("can-slider").then(function(CanSlider){

})
```
can you give the code example?

PR #1395, opened 2015-01-07
Adds autorender and can-import

justinbmeyer:
#1387 #1390 #1386

However, testee is freezing locally. @daffl can you fix this?


sykopomp:
I was working on this feature elsewhere, but I guess it's all yours now. See #1369. Nicer to use a parser anyway, I guess, and I like the explicit import tags and steal-based partial loading that PR supports. Would be nice to pull in that particular feature into here. Besides that, I guess may as well blow away that PR. The code was written based on the discussion at #1037.

I don't like automagic import any more than I would like magically trying to import can.Models based on their usage in JavaScript code: I think it's better to make it easy for people to specify modules to load, with explicit dependencies. But I don't really care about fighting about that. Can I just have a flag so I can disable that feature in my own projects?

justinbmeyer:
@sykopomp 

> I like the explicit import tags and steal-based partial loading that PR supports. 

I know what you mean by "partial loading", but what do you mean by "explicit import tags"?  There are explicit import tags: `<can-import>`

> I don't like automagic import any more than I would like magically trying to import can.Models based on their usage in JavaScript code

What do you mean?

justinbmeyer:
> I like the explicit import tags and steal-based partial loading that PR supports.

Actually, I'm not sure what you mean.  Do you mean my PR or yours?

> Nicer to use a parser anyway, I guess

There's nothing to guess about.  It's supporting an intermediate representation of the parsed output.  It haves the time to a "first" render.  Which will make production bundles faster.

daffl:
This is already in minor with the correct history.

PR #1397, opened 2015-01-11
All tests passing in IE8+ and evergreen browsers.

### LINKED ISSUE #1378: getOwnPropertyDescriptor causes fatal error in IE8
Within  can/construct/construct.js, this line will throw an error in IE8
```var descriptor = Object.getOwnPropertyDescriptor(newProps, name);```

This is because IE8 will only support getOwnPropertyDescriptor on DOM objects but not user defined objects.
http://msdn.microsoft.com/en-us/library/ie/dd548686(v=vs.94).aspx

Versions affected: 2.1.4

IE7 works

Additional comment: took out all js files apart from jQuery and can.custom.js to check.  Can.custom.js had been built with can.Map, which includes can.Construct.

daffl:
This pull request makes all tests pass in IE8+ and latest Chrome, Firefox, Safari and Opera. Closes #1378.


RunsFor:
Hey there! Great job!

Are there any plans to make `2.1.5` release in the nearest future with this pull request included?

We are using this framework in our daily work and really like it! Thanks.

PR #1412, opened 2015-01-21
#is helper added

### LINKED ISSUE #1406: {{#is}} and {{#eq}} helper
Are you considering adding stache/mustach ```{{#is}}``` helper in the core?

To allow view statement like this

```{{#is someAttr anotherAttr}} {{/is}}```

```{{^is someAttr 'someConstantValue' }} {{/is}}```

Having this is reducing the need of some exessive code to handle the view state, currenty it is done mainlty with {{if#}} helper,

{{#switch}} https://github.com/bitovi/canjs/issues/1264 migth be useful as well.

### PR DESCRIPTION
**Pull request by Alex0007**
Fixes #1406

### DIFF
--- a/view/stache/doc/helpers/is.md
+++ b/view/stache/doc/helpers/is.md
@@ -1,0 +1,24 @@
+@function can.stache.helpers.is {{#is expr1 expr2}}
+@parent can.stache.htags 12
+
+@signature `{{#is expr...}}BLOCK{{/is}}`
+
+Renders the `BLOCK` template within the current template.
+
+@param {can.stache.expression} [expr...] An expression or key that references a value within the current or parent
+
+@param {can.stache} BLOCK A template that is rendered
+if the result of comparsion `expr1` and `expr2` value is truthy.
+
+@return {DocumentFragment} If the key's value is truthy, the `BLOCK` is rendered with the
+current context and its value is returned; otherwise, an empty string.
+
+@body
+
+The `is` helper compares expr1 and expr2 and renders the blocks accordingly.
+
+	{{#is expr1 expr2}}
+		// truthy
+	{{else}}
+		// falsey
+	{{/is}}
--- a/view/mustache/doc/helpers.md
+++ b/view/mustache/doc/helpers.md
@@ -115,6 +115,28 @@ the current context so that all tags inside will look for keys on the local cont
 		<h1>Hi Austin</h1>
 		<p>You have a new friend: Justin</p>
 
+When using the `[can.mustache.helpers.is {{#is expr1 expr2}}]` helper you can simply compare 
+results of passed arguments. If the result of comparsion is **truthy**, the section will be rendered.
+
+	Template: 
+		<ul>
+			{{#is name 'Alex'}}
+				</li>Your name is {{name}}</li>
+			{{else}}
+				<li>Your name is not Alex!</li>
+			{{/if}}
+		</ul>
+
+	Data: 
+		{
+			name: 'John'
+		}
+
+	Result:
+		<ul>
+			<li>Your name is not Alex!</li>
+		</ul>
+
 The `[can.mustache.helpers.elementCallback {{(el)->CODE}}]` helper is a special helper for element callbacks that 
 will pass the active DOM element within the template to a function made up of inline code. This 
 is most useful for tasks such as initializing a jQuery plugin on the new HTML.
--- a/view/mustache/mustache.js
+++ b/view/mustache/mustache.js
@@ -1844,6 +1844,58 @@ steal('can/util',
 					return options.inverse(options.contexts || this);
 				}
 			},
+			/**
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: view/stache/doc/helpers/is.md [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+@signature `{{#is expr1 expr2}}BLOCK{{/is}}`
+
+Renders the `BLOCK` template within the current template.
+
+@param {can.stache.key} [args] A key that references a value within the current or parent
+context. If the value is a function or can.compute, the function's return value is used.
+
+@param {can.stache} BLOCK A stache template.
+
+@return {String} If the key's value is truthy, the `BLOCK` is rendered with the
```
justinbmeyer: instead of {String} this will return a {DocumentFragment}.

view/stache/doc/helpers/is.md
```diff
@@ -0,0 +1,27 @@
...
+current context and its value is returned; otherwise, an empty string.
+
+@param {can.stache} BLOCK A template that is rendered
```
justinbmeyer: Typically we always have all params before the @return.

view/stache/doc/helpers/is.md
```diff
@@ -0,0 +1,27 @@
+@function can.stache.helpers.is {{#is expr1 expr2 expr3}}
```
justinbmeyer: if you look at how the other helpers look in the sidebar of:

http://canjs.com/docs/can.stache.html

You'll notice that `{{#is expr1 expr2 expr3}}` doesn't really fit in.  At least, I think expr3 should be removed: `{{#is expr1 expr2}}`.  We could start showing expanded arguments like `{{#is expr...}}`

view/stache/doc/helpers/is.md
justinbmeyer: Actually, is this a duplicate?  I think one of the two BLOCK params should be removed.

view/stache/doc/helpers/is.md
```diff
@@ -0,0 +1,27 @@
...
+@parent can.stache.htags 12
+
+@signature `{{#is expr1 expr2}}BLOCK{{/is}}`
```
justinbmeyer: On pages like http://canjs.com/docs/can.stache.helpers.helper.html we use "..." to signal unlimited number of arguments.

Please change the signature to:

`{{#is expr...}}BLOCK{{/is}}`

view/stache/doc/helpers/is.md
```diff
@@ -0,0 +1,27 @@
...
+
+Renders the `BLOCK` template within the current template.
+
+@param {can.stache.key} [args] A key that references a value within the current or parent
```
justinbmeyer: We don't unfortunately have a can.stache.expression that you can link to.  However, I can fill that in.  Can you change this to:

```
@param {can.stache.expression} [expr...] An expression or key that references a value within the current or parent
```

justinbmeyer:
Looks very good.  I have some documentation suggestions.  Once those are fixed we can get this in.  Thanks!

whitecolor:
@justinbmeyer have looked at the code carefully? it assumes that this helper can take more then two params, is it ok with your vision of it?

justinbmeyer:
Yeah, I'm totally fine with it making sure all arguments are the same.  However, I don't think 2 loops are necessary.  

And, I'm unclear on this part:

```js
  if (can.isFunction(expr)) {
  	curValue = can.compute(expr)();
```

Normally, any function that is passed would be converted to a compute already.  There's no reason to do this afaik.

I would think something like this would work:

```js
var lastValue, curValue;
if(arguments.length - 2 <= 0 ){
  return options.inverse()
}

for (var i = 0; i < arguments.length - 2; i++) {
  var curValue = resolve(arguments[i]);
  if(i >= 0) {
    if(curValue !== lastValue) {
      return options.inverse();
    }
  }
  lastValue = curValue;
 }

return options.fn()
```

gsmeets:
dangit, now I'll have to refactor all my #eq's to #is's. ;)

daffl:
We could alias it. I think I have some `eq`s somewhere, too.

whitecolor:
@justinbmeyer 
Do we need to add ```eq``` alias for this helper to this pull?

daffl:
Thanks! This looks good and will make it into 2.2 with two things:
- Can you squash your commits (http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)
- Let's add the `eq` alias as well.

daffl:
Sweet! Thanks a lot, this is a great feature!

gsmeets:
Good job! And thanks for the alias. That saves me some work.

justinbmeyer:
One thing I noticed is that this isn't using `can.compute.truthy`, so it's possible that there might be unnecessary renderings of something like:

```
{{eq foo.length bar.length}}
```

If foo.length and bar.length were to change together in a batch, and stay the same, this helper would be called twice.

cc @akagomez

PR #1413, opened 2015-01-22
Fixed can.Object.same method

ablackthorn:
Since (typeof null === 'object'), can.Object.same failed to detect changes when a value is changed to null when the deep parameter is set to false.  This is the case when isDirty() is called from can.Map.backup.


James0x57:
Thank you for submitting this fix, unfortunately we were unable to reproduce the issue in the latest version.
We tested it in the backup plugin and in util object with both shallow and nested properties.

If there is still an issue, we would be happy to accept a pull request with tests!

Thank you,
//James

PR #1427, opened 2015-02-06
Fix initial value when using can-value with <select multiple>

### LINKED ISSUE #1414: CanJS can-value does not work for multi-select when options are rendered using Map
There seems to be an issue with Multi-selects when rendering using can-value.

Here is a link to the issue:

http://stackoverflow.com/questions/27759924/canjs-can-value-does-not-work-for-multi-select-when-options-are-rendered-using-m

asavoy:
Fixes #1414.

I just applied the solution that `Value.init()` uses for `<select>`, into `Multiselect.init()` so it works for `<select multiple>`.

See http://jsbin.com/bunezurubu/1/edit?html,js,output for an example of the bug.


web-mech:
:+1:

PR #1428, opened 2015-02-07
add/remove events always give you back arrays

### LINKED ISSUE #998: Fix `can.List` event docs, as well as an event ambiguity
### Summary

The `can.List` docs for event binding seem to be wrong about the `removedElements` value we actually get, and the `add` event actually returns a single value sometimes, which creates an ambiguity when listening to that event. Finally, `.pop()`ing an empty `can.List` gets you `[undefined]` as the `removedElements` value, which also creates an ambiguity.

#### Documentation corrections

The [docs for `can.List`](http://canjs.com/docs/can.List.prototype.attr.html#section_Events) say:

(for the `add` event)
```
* newElements - The new elements. If more than one element is added, newElements will be an array. Otherwise, it is simply the new element itself.
```

(for the `remove` event)
```
* removedElements - The removed elements. If more than one element was removed, removedElements will be an array. Otherwise, it is simply the element itself.
```

I could not find a case for the latter, since `can.List#removeAttr()` gives an empty array as `removedElements`, which is a pretty reasonable value. If that really is the case, the `remove` event docs should be updated to reflect this, and possibly even point out that you get an empty array back.

#### Add event is ambiguous for out-of-bounds `.attr()` calls

For the `add` event, I couldn't find any regular `can.List` methods that triggered a `newElements` that wasn't an array of the actual values, except for `can.List#attr`, which does return the single item added. (**EDIT**: I forgot to mention that this is **only for out-of-bounds indices**) This creates a pretty problematic ambiguity, since users will have no way of distinguishing an `add` event where multiple elements were added from an event where a single array was added at a single index. Unless I'm missing something, this should probably be fixed, and the docs updated accordingly to note how `newElements` will always be an array.

#### Remove event is ambiguous for `.pop()` on empty lists

(**EDIT**: I found this one not long after the above)
For the `remove` event, calling `can.List#pop()` on an empty list gives us the following information for the `change` and `remove` events:

* `change` - index: `"0"`, newVal: `undefined`, oldVal: `[undefined]`
* `remove` - index: `0`, newVal: `[undefined]`, oldVal: N/A (no such thing)

But if I have `new can.List([undefined])`, and call `.pop()` on that list, I get the exact same result.

This ambiguity is not present if you call `.splice(0, 1)` on both the `[undefined]` and empty lists.

#### Change events return index as string instead of integer

The `add`, `set`, and `remove` events all return their index as an integer, but the `change` event still uses a string. While we do want to (I assume) support `.attr()` on non-numberical properties of lists (since they kinda also work like objects/maps), I think we should emulate the behavior of `Arrays` and treat integer strings as integers, as well as in the interest of not surprising some poor soul who decides to switch events and suddenly gets `"12"` instead of `3`.

zkat:
can.List add event should always be an array

Don't trigger remove events on shift/pop for 0-length lists

For add events, include leading undefineds in the 'added' array

Fixes #998

### DIFF
--- a/list/list.js
+++ b/list/list.js
@@ -115,7 +115,7 @@ steal("can/util", "can/map", "can/map/bubble.js",function (can, Map, bubble) {
 				// `batchTrigger` direct add and remove events...
 				var index = +attr;
 				// Make sure this is not nested and not an expando
-				if (!~attr.indexOf('.') && !isNaN(index)) {
+				if (!~(""+attr).indexOf('.') && !isNaN(index)) {
 
 					if (how === 'add') {
 						can.batch.trigger(this, how, [newVal, index]);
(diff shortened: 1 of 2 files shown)


daffl:
Would this be a 3.0 change? I know it's not tested anywhere but wouldn't this break all code that's listening to an `add` event under the - admittedly wrong - assumption that you only get a single value?

justinbmeyer:
We document add always getting an array.  Our code was broken.  This is a bug fix.

PR #1429, opened 2015-02-07
can.Component now supports lexical semantics for <content>

### LINKED ISSUE #1069: Lexical semantics for can.Component
`can.Component` should have lexical semantics for its light dom. Consider this fiddle: http://jsfiddle.net/5D4FK/2/

Lexical semantics would shield the user from internal implementation details of the component. Likewise, they would shield the Component itself from unintended (or even intended) injection from the outside. Conceptually, the user should never know/care about the internal details of a component's shadow dom and interact with it purely through its exposed API (attributes and the light DOM, rendered in the lexical template context).

zkat:
Fixes #1069

### DIFF
--- a/component/component.md
+++ b/component/component.md
@@ -263,6 +263,50 @@ only renders friendly messages:
       }
     });
 
+### LeakScope
+
+A component's [can.Component::leakScope leakScope] option controls whether
+the surrounding template where the component is used can directly share
+scope with the component's internal scope. This option defaults to `true`,
+which means `<content>` is able to access internal scope variables on the
+component, and the component will be able to use outside variables when
+used on a page.
+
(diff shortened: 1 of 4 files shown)


daffl:
This is backwards compatible right? Can you rebase and double check why the tests are failing?

zkat:
@daffl this is up to date and clear.

component/component.md
```diff
@@ -263,6 +263,50 @@ only renders friendly messages:
       }
     });
 
+### LeakScope
```
justinbmeyer: I think leakScope should have it's own page, similar to how all the other prototype properties do.

justinbmeyer: Perhaps you can just move this into it's own MD file.

PR #1433, opened 2015-02-10
Updated: Ensure can-value bound input stay in sync with compute #888

cklanac:
I merged and resolved conflict from Matthew's original PR https://github.com/bitovi/canjs/pull/888. And updated setter based on Justin's reco here: https://github.com/bitovi/canjs/pull/888#discussion-diff-24426976

Matthew's PR fixes issue #887: can-value becomes out of sync when a compute rejects the new value #887


daffl:
This needs to be submitted against the minor branch. You can change the target branch when creating the PR so we'll probably have to close this one and submit it again.

PR #1440, opened 2015-02-12
Migrate to new Steal, CommonJS build, use steal-qunit

daffl:
This pull request migrates to the latest NPM version of Steal and steal-tools with all builds working, adds a CommonJS build and updates all tests and test pages to use steal-qunit.


daffl:
PS: This is using PhantomJS as the test runner.

PR #1444, opened 2015-02-13
Change attr to lowercase when setting attribute on element

### LINKED ISSUE #1201: {{prop}}CHECKED{{/prop}} doesn't work if someone manually changes input checked and then changes value elsewhere  (visible in TODOMVC)

cklanac:
Fixes #1201

### DIFF
--- a/util/attr/attr_test.js
+++ b/util/attr/attr_test.js
@@ -65,6 +65,25 @@ steal('can/util', 'can/util/attr', function () {
 		compute("bar");
 
 	});
+	
+	
+	/*test("attr.set CHECKED attribute works", function(){
+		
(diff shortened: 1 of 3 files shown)


util/attr/attr_test.js
```diff
@@ -65,6 +65,25 @@ steal('can/util', 'can/util/attr', function () {
 		compute("bar");
 
 	});
+	
+	
+	/*test("attr.set CHECKED attribute works", function(){
```
daffl: Why is this commented out?

cklanac:
Sorry, I pushed before committing the changes. I'll commit and restest and push again

PR #1451, opened 2015-02-14
Fix #1416

### LINKED ISSUE #1416: Define get: async arguments are sometimes undefined
I'm using the `can/map/define` plugin, specifically the async form of `get:`. There's not any docs on this, despite  there being examples for it, such as [canjsmeetup/1_intro](https://github.com/bitovi/canjsmeetup/blob/dba08dd71603dd0b13a1a65e6fb3e7322a44d15a/1_intro/two_one_example.html) and [canjsmeetup/7_route](https://github.com/bitovi/canjsmeetup/blob/dba08dd71603dd0b13a1a65e6fb3e7322a44d15a/7_route/components/contact.js), so I'm not sure what is exactly supported.

The issue is that the callback argument is undefined, *unless* something has bound to the property first.

Please see this [JSBin](http://jsbin.com/wotititozo/1/edit?js,console), it errors at (1), but it works at (2) after doing `person.bind('fullName', function() {})`.

asavoy:
Proposed fix for #1416, plus a test case.

I'm open to feedback on this, as I'm not super familiar with the internals of observables.

### DIFF
--- a/map/define/define.js
+++ b/map/define/define.js
@@ -158,6 +158,22 @@ steal('can/util', 'can/observe', function (can) {
 		return this;
 	};
 
+	var oldGet = proto.__get;
+	proto.__get = function (attr) {
+		if (attr) {
+			// If it's an async getter and unbound, bind to it so that we don't
+			// raise an error within the compute. See #1416.
+			var computedBindings = this._computedBindings;
+			var def = this.define;
+			var isAsyncGetter = (def && def[attr] && def[attr].get && def[attr].get.length === 2);
+			if (isAsyncGetter && !computedBindings[attr].count) {
+				this[attr].bind('change', function() {});
+				computedBindings[attr].count += 1;
+			}
+		}
+		return oldGet.call(this, attr);
+	};
+
 	var converters = {
 		'date': function (str) {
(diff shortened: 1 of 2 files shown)


map/define/define.js
```diff
@@ -158,6 +158,22 @@ steal('can/util', 'can/observe', function (can) {
...
+	var oldGet = proto.__get;
+	proto.__get = function (attr) {
+		if (attr) {
+			// If it's an async getter and unbound, bind to it so that we don't
+			// raise an error within the compute. See #1416.
+			var computedBindings = this._computedBindings;
+			var def = this.define;
+			var isAsyncGetter = (def && def[attr] && def[attr].get && def[attr].get.length === 2);
+			if (isAsyncGetter && !computedBindings[attr].count) {
+				this[attr].bind('change', function() {});
```
daffl: I'm not sure but wouldn't this leak? Do we need to unbind when the getter returns?

justinbmeyer:
Thanks @asavoy! I'm closing as I think this will leak.   It's documented in 3.0 can-define that a getter called without binding will not be passed a `resolve` argument.

Issue #828, opened 2014-03-26
can.list.sort triggers event before sorting but none after


initialize:
Suggestion: 
https://github.com/initialize/canjs/commit/2c1ef1faf24df959a948338be8daf2939f1f02bd

(Sorry, I'm kinda new to github. Is this the right way to suggest changes?)

daffl:
This has been fixed via #1454

PR #1454, opened 2015-02-16
Complete the Sort plugin

### LINKED ISSUE #1404: can.List.prototype.sortIndexes is defined but not referenced
`sortIndexes` is defined here: 

https://github.com/bitovi/canjs/blob/8425837c343efd6828bba1040bdabdd26ab96bea/map/sort/sort.js#L31

But a search yields only one result: 

```bash
totes:canjs akagomez$ grep -rn "sortIndexes" .
./map/sort/sort.js:31:		sortIndexes: [],
```

It's possible that it's referenced using some funky string concatenation, but that seems unlikely to me since `sort.js` itself doesn't utilize it.

akagomez:
As far as I could tell the Sort plugin was largely incomplete. 

After a lot of work it now...

- Two-way binds a `can.List`'s sort order to the DOM
- Makes the fewest amount of DOM manipulations possible when rendered with `{{#each}}`
- Ensures that operations like `push`, `unshift`, and `splice` add items at the correct index via `can.List._getInsertIndex()` (for efficiency there's no need to _append_ then _move_ an item)
- Operations that result in an item being moved from one index to another result in a `move` event (this used to result in a `remove` and `add` event)
- DOM manipulations made on behalf of the `{{#each}}` helper _move_ the underlying DOM nodes rather than _removing_ and _re-creating_ them (this is much faster).
- Resolves index changes to individual items with  `can.List._getInsertIndex` as opposed to a complete `can.List.sort` operation
- Changes to items in the `can.List` that have a `batchNum` result in a complete `can.List.sort` operation rather than (potentially) _lots_ of far less efficient `can.List._getInsertIndex` operations
- Breaks from `can.List.sort` operations when a partial sort is all that's needed to resolve the entire `can.List`'s sort order 

Future goals: 

- ~~Rough out documentation~~
- Choose a more efficient sorting strategy when there are no `{{#each}}` dependencies
- Benchmark this implementation of the Sort plugin against https://github.com/Matt-Esch/virtual-dom

Closes #1404, #1114, #1265, #828


asavoy:
+1 This looks awesome

imjoshdean:
:+1:

matthewp:
This might need to be redirect to minor. @daffl is that right?

daffl:
No. Master is up to date and ready for 2.2. Ideally things go against master now but it doesn't really matter since I need to merge minor in again anyway.

marshallswain:
If you can sneak it into master, that would be awesome.

PR #1455, opened 2015-02-18
Added jquery to peer dependencies

mjstahl:
Changed the verion of the required jquery to a range of 1.9.0 to less than 3.0.0

Kept jquery in dev dependencies so that it can be install at the root level of canjs via npm install

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
 		"url": "git@github.com:bitovi/canjs.git",
 		"github": "https://github.com/bitovi/canjs"
 	},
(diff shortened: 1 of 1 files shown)


mjstahl:
jquery is now in both peerDependencies and devDependencies.  

It was originally thought that only having it in peerDependencies would be enough, but for versions of npm (such as 1.4.28 -- what we test with in Travis CI) peerDependencies do not get installed when doing an "npm install" at the root level of the project.

PR #1458, opened 2015-02-19
Add check in live.list to determine if current event is part of a batch.

### LINKED ISSUE #680: can.batch.start/stop causes can.List with length>1 to render last element twice
```handlebars
{{#each arr}} {{.}} {{/each}}
```
```js
for (var i=1; i<=2; i++) {
    this.arr.push(i);
}
```
will produce `1,2`, however:
```js
can.batch.start();
for (var i=1; i<=2; i++) {
    this.arr.push(i);
}
can.batch.stop();
```
will produce `1,2,2`.

?

http://jsfiddle.net/prometh/34YZP/

cklanac:
Add a batchNum property to can.batch. Add check in can.live.list add method to determine if current event was triggered during a batch job. if so, then exit the add. Fixes #680.

### DIFF
--- a/util/batch/batch.js
+++ b/util/batch/batch.js
@@ -173,6 +173,8 @@ steal('can/util/can.js', function (can) {
 					i, len;
 				batchEvents = [];
 				stopCallbacks = [];
+				// Capture current batchNum so it can be check in live bindings
+				can.batch.batchNum = batchNum;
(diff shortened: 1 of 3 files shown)


daffl:
Test?

daffl:
:+1: once tests are in.

PR #1462, opened 2015-02-20
Breaking Test to ensure that model is removed from store on ID change

cklanac:
Breaking Test to accompany issue #1463.

This is a (currently) breaking test to check that a model is removed from the store if the ID changes. Typically the ID shouldn't change during the life of a model. But under some circumstances such as creating/saving - the ID might not be known on creation and only resolved after saving. This test should ensure that the old model is removed when the ID is updated/changed.

### DIFF
--- a/model/model_test.js
+++ b/model/model_test.js
@@ -1748,5 +1748,19 @@ steal("can/model", 'can/map/attributes', "can/test", "can/util/fixture", functio
 
 	});
 
+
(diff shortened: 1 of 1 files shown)


daffl:
You probably need to pull in latest master.

cklanac:
This is just a breaking test to go along with the Issue... no a fix

Issue #1447, opened 2015-02-13
Add Dojo tests and can/view/autorender


daffl:
Currently the Dojo tests and can/view/autorender plugin are not included in the full test suite. Dojo tests need a build update to not overwrite Dojo's define and the autorender tests are failing in PhantomJS only.

matthewp:
isn't autorender added now?

daffl:
Oh yes, both is. We can close this.

PR #1466, opened 2015-02-23
Dojo library should require Deferred

matthewp:
Dojo requires its dependencies. At some point the file was modified and
it requires `../deferred` instead of `../Deferred` (as it does in the
		Dojo source) which caused the dist tests to break. Fixing the
capitalization issue fixes it. Part of #1447


matthewp:
Going to try including a newer version of dojo. `1.10.4`

Issue #530, opened 2013-11-07
Wrapping can.compute in can.Map breaks live-binding


ghost:
When a can.compute is wrapped in a can.Map and then live-bound, the view will not update when the compute changes.
I have an app that has a global can.Map for storing the state, so when I now put in a can.compute anywhere in this Map (manually or by accessing can.route.data for example, which is part of this state), the compute will never update the view.

See here for a very basic example: http://jsfiddle.net/simpleFabian/FLj7b

Is this behaviour intended in some way? 
Can I work around it temporarily in any way? (as in using a helper, maybe?)

Thanks in advance!

PR #1470, opened 2015-02-25
Add a "compute" type converter to the Define plugin

### LINKED ISSUE #1409: Insert can.compute into can.Map
Hello,

One feature I would really appreciate - insert can.compute (the value one, not the function one) into can.Map, so I can use it with normal can.Map interface and having all bindings working on 'both sides', as it works when inserting can.Map into other can.Map, example:

```javascript
var slaveMap = new can.Map({foo: 1});
var masterMap = new can.Map(innerMap: slaveMap);

// then this works
masterMap.innerMap.attr('foo', 5);
slaveMap.attr('foo') === 5; // true


var compute = can.compute(1);
var map = new can.Map(insertedCompute: compute);

// how it works
map.insertedCompute(); // yields 1
map.insertedCompute(5); // sets 5

// how it would be awesome
map.attr('insertedCompute') // yields 1
map.attr('insertedCompute', 5) // sets 5
```

What is the main goal? To get this pattern working:
```javascript
var map1 = new can.Map({a: 1, b: 2});
var map2 = new can.Map({c: 3, d: 4});

var map3 = new can.Map({
    a: map1.compute('a'),
    c: map2.compute('c')
});
```

Since Components are based on data coupling it would be extremely useful to have this possibility.

Unfortunately, there would be problem with backward compatibility, since someone could use it nowadays with current functionality, but it could be done by defining method like:

```javascript
var compute = can.compute(1);
var map = new can.Map({});

map.insertCompute('nameOfAttr', compute);
```

akagomez:
Currently you can set/get a `can.compute`'s value with `can.Map.prototype.attr` if it is defined as a property of the map's prototype. Example: 

```
var MapWithCompute = can.Map.extend({
  computed: can.compute(0, function (newVal, oldVal) {
      if (typeof newVal !== 'undefined') {
          return newVal; 
      }
      return oldVal; 
  })
}); 

var m = new MapWithCompute({
    computed: 1
}); 
console.log(m.attr('computed')); // -> 1

m.attr('computed', 2); 

console.log(m.attr('computed')); // -> 2
```
http://jsfiddle.net/akagomez/90ottotb/

However, if you set the value of a property on a `can.Map` to a `can.compute` *using* `can.Map.prototype.attr`, the API for setting/getting that `can.computes` value changes. Example: 

```
var MapWithCompute = can.Map.extend({}); 

var m = new MapWithCompute();
m.attr({
    computed: can.compute(0, function (newVal, oldVal) {
        if (typeof newVal !== 'undefined') {
            return newVal; 
        }
        return oldVal; 
    })
}); 
console.log(typeof m.attr('computed')); // -> function
console.log(m.attr('computed')()); // -> 0

m.attr('computed', 2); 

console.log(m.attr('computed')); // -> 2
console.log(typeof m.attr('computed')); // -> 'number'
```
http://jsfiddle.net/akagomez/jt9x7kmr/

This is due to the fact that `can.Map.prototype.attr`sets/gets the value explicitly, per #530. 

By adding a "compute" type to the Define plugin (and extending `can.Map.prototype.___set`) I was able to process `can.compute`'s that are set with `can.Map.prototype.attr` in the same way that `can.computes` are processed on the prototype with `can.Map.prototype.setup`. 

This enables us to do things like this: 

```
var ToggleMap = can.Map.extend({
	define: {
		computable: {
			type: 'compute'
		}
	}
});

var toggleCompute = can.compute(false);

var sharedToggle1 = new ToggleMap({
	computable: toggleCompute
});

var sharedToggle2 = new ToggleMap({
	computable: toggleCompute
});

var soloToggle = new ToggleMap({
	computable: false // Converted to can.compute
});

console.log(sharedToggle1.attr('computable')); // -> false
console.log(sharedToggle2.attr('computable')); // -> false
console.log(soloToggle.attr('computable')); // -> false

// Update a reference to the compute
sharedToggle1.attr('computable', true);

console.log(sharedToggle1.attr('computable')); // -> true
console.log(sharedToggle2.attr('computable')); // -> true
console.log(soloToggle.attr('computable')); // -> false

// Update a reference to the compute
sharedToggle2.attr('computable', false);

console.log(sharedToggle1.attr('computable')); // -> false
console.log(sharedToggle2.attr('computable')); // -> false
console.log(soloToggle.attr('computable')); // -> false

// Update a reference to the compute
toggleCompute(true);

console.log(sharedToggle1.attr('computable')); // -> true
console.log(sharedToggle2.attr('computable')); // -> true
console.log(soloToggle.attr('computable')); // -> false
``` 

Closes #1409

### DIFF
--- a/map/define/define.js
+++ b/map/define/define.js
@@ -43,9 +43,9 @@ steal('can/util', 'can/observe', function (can) {
 			Map = this.constructor,
 			originalGet = this._get;
 
-		// Overwrite this._get with a version that commits defaults to
-		// this.attr() as needed. Because calling this.attr() for each individual
-		// default would be expensive.
+		// Temporarily overwrite this._get with a version that commits defaults to
+		// this.attr() as they are used in other generated defaults. Because
+		// calling this.attr() for each individual default would be expensive.
 		this._get = function (originalProp) {
 
 			// If a this.attr() was called using dot syntax (e.g number.0),
@@ -63,7 +63,7 @@ steal('can/util', 'can/observe', function (can) {
 				// other defaultGenerators.
 				this.attr(prop, defaults[prop]);
 
-				// Make not so that we don't commit this property again.
+				// Make note so that we don't commit this property again.
 				propsCommittedToAttr[prop] = true;
 			}
 
@@ -86,7 +86,7 @@ steal('can/util', 'can/observe', function (can) {
 
 
 	var proto = can.Map.prototype,
-		oldSet = proto.__set;
+		__set = proto.__set;
 	proto.__set = function (prop, value, current, success, error) {
 		//!steal-remove-start
 		var asyncTimer;
@@ -124,7 +124,7 @@ steal('can/util', 'can/observe', function (can) {
 			var setterCalled = false,
 
 				setValue = setter.call(this, value, function (value) {
-					oldSet.call(self, prop, value, current, success, errorCallback);
+					__set.call(self, prop, value, current, success, errorCallback);
 					setterCalled = true;
 					//!steal-remove-start
 					clearTimeout(asyncTimer);
@@ -146,7 +146,7 @@ steal('can/util', 'can/observe', function (can) {
 				return;
 			} else {
 				if (!setterCalled) {
-					oldSet.call(self, prop,
+					__set.call(self, prop,
 						// if no arguments, we are side-effects only
 						setter.length === 0 && setValue === undefined ? value : setValue,
 						current,
@@ -158,10 +158,30 @@ steal('can/util', 'can/observe', function (can) {
 			}
 
 		} else {
-			oldSet.call(self, prop, value, current, success, errorCallback);
+			__set.call(self, prop, value, current, success, errorCallback);
 		}
 
 		return this;
+	};
+
+	var ___set = proto.___set;
+	proto.___set = function (prop, val) {
+		var type = getPropDefineBehavior("type", prop, this.define);
+
+		// If this property's "type" is not "compute", set the value as always
+		if (type !== 'compute' || ! val.isComputed) {
+			return ___set.apply(this, arguments);
+		}
+
+		// If this property's "type" is "compute", treat the compute like
+		// a can.Map does on setup
+		//   1) Save the compute so that it can be written to in the original
+		//      ___set
+		this[prop] = val;
+		//   2) Track the number of uses
+		this._trackCompute(prop);
+
+		return ___set.call(this, prop, val());
 	};
 
 	var converters = {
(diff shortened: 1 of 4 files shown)


daffl:
This is great... but... docs?

justinbmeyer:
Indeed, this will be super useful. I'm wondering if it's possible to implement without depending on _computedBindings.

akagomez:
I was a little annoyed at how sloppily I was testing this feature, so I cleaned that up a bit. 

Also, made it so that functions passed to `can.Map.prototype.attr` get converted to `can.compute`'s in the context of the `can.Map`. 

@daffl Documentation added. It's a bit more wordy than the other types in the type list, but I think it works (for now).  

@justinbmeyer I can't see any way around it due to `bind` and `unbind`'s dependence on `_computedBindings`. 

However, unless I'm missing something we can replace `_computedBindings` in `can.Map.prototype.setup`, `can.Map.prototype.__get`, and `can.Map.prototype.___set` with  `val.isComputed`. Then, account for the fact that the initial `_computedBindings.propertyName` object may not be defined yet in `bind` and `unbind`. They seem to be the only methods that depend on the binding count.  

All of that said, I think that should be handled in a separate PR since it works as it stands.

daffl:
Awesome, :+1: from me! I agree that the refactoring can be done in separately but can you maybe create a follow-up issue for it so that we don't loose track. Thanks!

justinbmeyer:
@akagomez Clean up is fine in this pull request as it may need to be cleaned before it gets added.  I'm going to look at it tonight.

justinbmeyer:
I added some tests for a define get/set that "should" provide the same behavior that this pull request is providing (but without the special type) here:

https://github.com/bitovi/canjs/commit/0e2ab190d95f14b5cf0936990dc1f43725f728b9

This exposes a few problems with `can.compute` that I'd like to address and should make this functionality easier to implement.

The primary problem is that I'd like to use only one compute to implement this functionality. That compute is the `compute.async` created by define.get.  

The "real" problem is that setting a `compute.async` doesn't do any actual setting.  There are a few places I've started to fix this:

Fix 1 - a setter should update the cached value when called.

```
                                        context = options.context || options;
                                        get = options.get || get;
-                                       set = options.set || function(){
-                                               return value;
+                                       set = options.set || function(newVal){
+                                               return value = newVal;
                                        };
                                        // 
```

Fix 2 - When a compute is set, it should not assume that the passed value is the new cached value.  The following makes it call the compute's function to get the actual cached value.

```
                                // If the computed function has dependencies,
                                // return the current value
                                if (computed.hasDependencies) {
-                                       return get.call(context);
+                                       return value = get.call(context);
                                }
```


Finally, even when setting a compute the following causes problems:

```js
if (computed.hasDependencies) {
   return value = get.call(context);
}
```


It calls the compute's function, but is not doing it in a way that re-evaluates the value and might trigger a change event.  

Ideally, the code within `setupComputeHandlers`'s `onchanged` should be rerun. A "set" should be treated similar to a dependent observable changing.

I'm thinking this work should be done in @alexisabril 's `can.Compute`

justinbmeyer:
`can.Compute` passes tests #1482

akagomez:
#1486 does this better by giving converters the ability to mixin `set` and `get` functions and reworks `can.compute` so that `get` can be derived from the result of `set`.

PR #1486, opened 2015-03-03
Adds can.Compute, and "compute" type and setter / getter paring for define pugin

justinbmeyer:
You can:

```js
c = new can.Compute(5);
c.get()
c.set(6)
c.bind("change", callback)
```

```js
MyMap = can.Map.extend({
  define: {value: {type: "compute"} }
})
var myMap = new MyMap();
var c = can.compute(5);

myMap.attr("value",c);
myMapp.attr("value") //-> 5

c(6)
myMapp.attr("value") //-> 6

var c2 = can.compute("a")
myMap.attr("value",c2);
myMapp.attr("value") //-> "a"
```

```js
MyMap = can.Map.extend({
  define: {
    num: {
      set: function(newValue){
        return +newValue;
      },
      get: function(lastSetValue){
        return Math.pow( lastSetValue, this.attr("pow") )
      }
    }
  }
})

var myMap = new MyMap({pow: 2});
myMap.attr("num","4");
myMap.attr("num") //-> 16
```


justinbmeyer:
@daffl I'm assigning this to you as this might break a few of the `compute.async` tests.  You can probably comment them out in the legacy tests b/c they aren't a documented API.

akagomez:
From the looks of it, this is now possible too:

```
can.define.powerOfTwo = {
  set: function(newValue){
    return +newValue;
  },
  get: function(lastSetValue){
    return Math.pow( lastSetValue, 2)
  }
}

MyMap = can.Map.extend({
  define: {
    num: {
      type: 'powerOfTwo'
    }
  }
})

var myMap = new MyMap();
myMap.attr("num","4");
myMap.attr("num") //-> 16
```

compute/compute.js
```diff
@@ -37,563 +37,36 @@ steal('can/util', 'can/util/bind', 'can/util/batch', function (can, bind) {
...
-		}
-	};
-	can.__addReading = function(o){
-		if (stack.length) {
-			can.simpleExtend(stack[stack.length-1], o);
-		}
-	};
+		compute.bind = can.proxy(internalCompute.bind, internalCompute);
+		compute.unbind = can.proxy(internalCompute.unbind, internalCompute);
+		compute.isComputed = can.proxy(internalCompute.isComputed, internalCompute);
```
daffl: How is this supposed to work? For some reason it is only breaking for Mootools but `can.proxy(true, internalCompute)` doesn't seem right.

justinbmeyer: isComputed is a boolean.  So I think this should be compute.isComputed = true

akagomez: In my playing with this branch I noticed that isComputed was no longer a `boolean` and was uber confused. Glad that's cleared up.

Issue #1361, opened 2014-12-08
Proposal: Argument support for `can-click`


zkat:
## Proposal

`can-click` handlers should support inline argument passing, right in the template. 

## Justification and Notes

This would obviate the need to entire categories of javascript-side viewmodel methods and allow us to add a ton of simple functionality right in the template. A common pattern in component-heavy application is writing relatively trivial methods that end up being thin wrappers over, say, `.attr()`.

I've also seen a common case where we have a loop, and then we do something like `<li can-click="{activateTab}" data-tab-name="@index">`, and then end up writing some JS that has to manually yank `tab-name` out of the element, and just do that dance a lot. We can do better just by allowing argument-passing.

## Example

Assuming #1360 is in place, the following would be a simple `tabs` component (collecting tabs array is left as an exercise to the reader, and might be potentially served by #500):

```html
<ul class="tabs">
  {{#each tabs}}
    <li can-click="{self.attr 'tabIndex' @index}">
      {{name}}
    </li>
  {{/each}}
</ul>
{{#each tabs}}
  {{#equal self.tabIndex @index}}
    {{{content}}} {{! or <content select="nth(@index)"></content> }}
  {{/each}}
{{/each}}
```

justinbmeyer:
This is a duplicate of #1219.  Can't agree more. But closing as that issue was first and includes ideas on how to pass the element and events objects.

Issue #1300, opened 2014-11-03
rename can.Component scope to context?


stevenvachon:
It was suggested way back in #635 and I agree with it.

`scope` may be good for those migrating from AngularJS but to everyone else, I think that `context` makes more sense. It also causes less confusion with can-EVENT function arguments:
```js
context: {
    canEventFunction: function(scope, element, event) {
        // scope vs context
    }
}
```

Maybe a 3.0 feature?

justinbmeyer:
Some definitions:

## In Mustache

- context - the initial object a template will look within for key values
- key - an identifier
- scope - a `can.view.Scope` that represents a context and a parent context.
- scope chain - all contexts a key might be found within for a certain area of a template

## In attribute helpers

- `can-scope` - a helper that binds a component's "scope" to a particular key value in the current context.

## In can.Component

- scope - A definition of a `can.Map` that is used to render the component's template.  This is normally used as a view model.

daffl:
Will be renamed to `viewModel` via #1512.

PR #1495, opened 2015-03-04
Argument support for can-EVENT

### LINKED ISSUE #1219: can-EVENT arguments

daffl:
can-EVENT now parses its arguments similar to how stache helpers handle it and prepends those arguments and the hash to the arg list.

Resubmission of #1357 see #1361 and closes #1219

### DIFF
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -3,7 +3,7 @@
 // This file defines the `can-value` attribute for two-way bindings and the `can-EVENT` attribute 
 // for in template event bindings. These are usable in any mustache template, but mainly and documented 
 // for use within can.Component.
-steal("can/util", "can/view/callbacks", "can/control", function (can) {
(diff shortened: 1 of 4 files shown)


view/bindings/bindings.js
```diff
@@ -176,26 +176,87 @@ steal("can/util", "can/view/callbacks", "can/control", function (can) {
...
+					return null;
 				}
+				//!steal-remove-end
+
+				var args = [];
+				var $el = can.$(this);
+				var localScope = data.scope.add({
+					"@element": $el,
+					"@event": ev,
+					"@scope": can.scope($el[0]),
```
justinbmeyer: @zkat or @daffl Can we remove this line for now, until #1300 is figured out. I'm pretty sure I'll want @scope to mean the actual `can.view.Scope` at that time.

PR #1504, opened 2015-03-10
Fix can.scope and autorender

### LINKED ISSUE #1503: autorender test failing
In latest master the autorender test is failing in Chrome and Firefox. I'm investigating what the issue is.

matthewp:
This fixes can.scope and autorender. There were two problems:

1. autorender depended on can.scope to create a scope for an element
when one didn't already exist and this no longer was happening. The fix
was to make can.scope create a scope if there isn't already one. Added a
test for can.scope for this.

2. autorender was setting up scope for the scoped script tag and adding
attributes to the scope but this wasn't working. The reason is that
`el.attributes` is a NamedNodeMap of Attrs which are objects, not
strings. The fix was to use the `attr.name` property.

Fixes #1503

### DIFF
--- a/builder.json
+++ b/builder.json
@@ -213,6 +213,11 @@
 			"type": "core",
 			"isDefault": true,
 			"hidden": true
+		},
+		"can/view/autorender/autorender": {
+			"name": "can.autorender",
+			"type": "plugin",
+			"description": "Automatically render templates found in the document"
 		}
 	},
 	"types": {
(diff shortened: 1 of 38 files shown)


daffl:
I also added the autorender module back into the builder.json but it won't run the tests (Phantom gets stuck on those for some convoluted reason and some random error somewhere in Underscore that I wasn't able to figure out yet).

justinbmeyer:
Underscore?!?

daffl:
In the test runner. Seems like the way the QUnit test are run in the iFrame it doesn't know what suite they belong to and thus gets stuck.

justinbmeyer:
maybe there's code to import from can.route's tests that use iframes.


Justin Meyer
847-924-6039

daffl:
Looks like I found a way. Will merge once the tests pass.

Issue #1510, opened 2015-03-11
Sort plugin fires a "reset" event for every item added with push/unshift


akagomez:
My gut tells me this line should be moved outside the `while` loop: 

```javascript
can.batch.trigger(this, 'reset', [args]);
```
https://github.com/bitovi/canjs/blob/6fb12f5f56be10aa49567345452a35117fb8d5f2/map/sort/sort.js#L255

daffl:
Closed via #1517

Issue #1509, opened 2015-03-11
Sort plugin fires "add" event incorrectly when .push() is called with multiple items


akagomez:
For each item that's added to a sorted `can.List` with `.push()` or `.unshift()`, an `add` event is fired with ALL of the items that were passed.

An example: 

```
list.bind('add', function (ev, items, index) {
  console.log('# of items:', items.length); 
}); 
list.push(['a', 'b', 'c']); // -> # of items: 3
                            // -> # of items: 3
                            // -> # of items: 3
```

akagomez:
I have a fix ready. I just need to create a test to go along with it.

daffl:
Closed via #1517

PR #1517, opened 2015-03-11
Sort plugin: Fix events; Fix demo; Relocate to can/list/sort

### LINKED ISSUE #1508: Sort plugin _bubbleRule breaks "destroyed" event
If I comment out this code, the "destroyed" event fires as expected when calling `.destroy()`... 

```javascript
	can.List._bubbleRule = function(eventName, list) {
		if(list.comparator) {
			return "change";
		}
		return oldBubbleRule.apply(this, arguments);
	};
```
https://github.com/bitovi/canjs/blob/6fb12f5f56be10aa49567345452a35117fb8d5f2/map/sort/sort.js#L5-L10

I'll need some help understanding what `_bubbleRule` is meant to do before I can work out a solution.

akagomez:
Closes #1508, #1509, #1510.


daffl:
You probably need to update https://github.com/bitovi/canjs/blob/master/builder.json#L154 if you changed the location.

Issue #1519, opened 2015-03-13
Problem with defines called out of order


moschel:
Defines that are dependent on each other seem to not fire in the correct order. I have a breaking test showing this here: https://github.com/bitovi/canjs/commit/a3d9d38915e5425808d94b677a092f5a0e7cb22f

moschel:
Here's the breaking test:

````js
		var MyMap = can.Map.extend({
		  define: {
		    fulfillmentType: {
		      get: function() {
		        return this.attr('ft');
		      },

		      set: function(newVal) {
		        this.attr('ft', newVal);
		        return newVal;
		      }
		    },

		    propA: {
		      get: function() {
		        return this.attr('fulfillmentType') == 'a';
		      }
		    },

		    propB: {
		      get: function() {
		        return this.attr('fulfillmentType') == 'b';
		      }
		    },
		    
		    combined: {
		      get: function() {
		        return this.attr('propA') || this.attr('propB');
		      }
		    }
		  }
		});

		var map = new MyMap({ft: 'a'});

		map.on('combined', function(){});
		map.on('propA', function(){});
		map.on('propB', function(){});

		map.attr('fulfillmentType', 'b');
		equal(map.attr('combined'), true);
````

This seems very dependent on the order. If I change the order the properties are bound, it doesn't break. If I remove propA or propB altogether, it doesn't break.

daffl:
This is closed via #1520

m-mujica:
It seems that this issue was not totally fixed. In [this branch](https://github.com/bitovi/canjs/tree/bug-nested-computes) I've created a new failing test with more complex computes. 

![nested-computes](https://cloud.githubusercontent.com/assets/724877/7068077/47f5b7b8-dea6-11e4-8683-319eb8ed52ce.png)

I've also created a [jsbin](http://jsbin.com/limobe/2/edit?js,console) with application like code that this issue affects.

PR #1520, opened 2015-03-13
makes events triggered for the current batch within the current batch ha...

justinbmeyer:
...ppen after all schedule event handlers are fired #1519

### DIFF
--- a/compute/compute_test.js
+++ b/compute/compute_test.js
@@ -749,7 +749,6 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
 		
 		equal(combined(), true);
 		
-		console.log('binding to combined');
 		combined.bind('change', function(){ });
 		
 		can.batch.start();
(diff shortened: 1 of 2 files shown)


compute/compute_test.js
```diff
@@ -727,5 +727,37 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
...
+		var combined = can.compute(function(){
+			var valA = propA(),
+				valB = propB();
+	
+			return valA || valB;
+		});
+		
+		equal(combined(), true);
+		
+		console.log('binding to combined');
```
m-mujica: you forgot to remove this (I think?)

justinbmeyer: arg, yes.  Can you or @brianmoschel remove it for me?

m-mujica: yeah, I'll do that right away.

PR #1522, opened 2015-03-14
Backport compute set value for input type radio and checkbox

### LINKED ISSUE #1518: Keep can-value in sync with compute for input type="radio/checkbox/select"
We did this for `<input type="text">` here: #1443 

I'd hate for the one to be fixed in 2.2, and the others not. 

Just this morning @imjoshdean tried to rely on this and got burned: http://jsfiddle.net/imjoshdean/npwpa6v9/

I think  that as the Define plugin gets more popular this will become a greater issue. 

@daffl @justinbmeyer, Do you think this is important enough to hold up the 2.2 release?

akagomez:
Closes #1518 

- Passed a reference to the radio input's parent compute (I named it `abstractValue`) to the `Checked` control.
- Added a `triggerWhileUnchanged` option to the event object passed to `can.dispatch` and subsequently `onchanged` to coerce the child computes to trigger `change` events despite their values not changing.
- If a radio input is selected and the value that's passed to the setter doesn't stick, I trigger a change event on the `abstractValue` and set `triggerWhileUnchanged` to true so that each radio input can read the actual value and set itself to the selected/not-selected state.
- In the case of a "checkbox" type input, we just re-run that input's `check` function again so that it can detect the actual value
- `<select>` inputs already handled the persisted compute value correctly. I just added a test.

### DIFF
--- a/compute/proto_compute.js
+++ b/compute/proto_compute.js
@@ -79,7 +79,7 @@ steal('can/util', 'can/util/bind', 'can/util/batch', function (can, bind) {
 
 	var updateOnChange = function(compute, newValue, oldValue, batchNum){
 		// Only trigger event when value has changed
-		if (newValue !== oldValue) {
+		if (newValue !== oldValue || compute._triggerOnSameValue) {
 			can.batch.trigger(compute, batchNum ? {type: "change", batchNum: batchNum} : 'change', [
 				newValue,
 				oldValue
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -102,10 +102,18 @@ steal("can/util", "can/view/stache/mustache_core.js", "can/view/callbacks", "can
 			}
 
 			if (el.type === "checkbox" || el.type === "radio") {
+
+				// Configure this compute trigger change events even when its dependent
+				// compute triggers a change event after having not changed. This
+				// is so that each radio instance can synchronize itself with the
+				// *actual* set value after user interaction.
+				value.computeInstance._triggerOnSameValue = true;
+
 				// For checkboxes and radio buttons, create a Checked can.Control around the input.  Pass in 
 				// the compute representing the can-value and can-true-value and can-false-value properties (if 
 				// they were used).
 				new Checked(el, {
+					abstractValue: data.scope._context[attr],
 					value: value,
 					trueValue: trueValue,
 					falseValue: falseValue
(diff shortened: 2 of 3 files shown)


akagomez:
@justinbmeyer I'd love it if you could take a look and let me know if there's a better way to accomplish this.

akagomez:
After looking at this again I've decided that the current `compute._triggerWhileUnchanged` approach will not work since it persists once set. 

I think the solution is to pass the event object to `updater` and `updateOnChange`, then add some logic to `updateOnChange` that looks for `eventObject.ignoreSameValues === false`, or something like that.

compute/proto_compute.js
```diff
@@ -77,10 +77,22 @@ steal('can/util', 'can/util/bind', 'can/util/batch', function (can, bind) {
...
+		if (ev) {
+			if (ev.batchNum) {
+				newEv.batchNum = ev.batchNum;
+			}
+
+			if (ev.triggerOnSameValue) {
+				newEv.triggerOnSameValue = ev.triggerOnSameValue;
+			}
+		}
+
```
akagomez: There's probably a cleaner way to write this. I didn't spend much time on it.

akagomez:
Thinking about it a bit more, this approach... 

```
// The value didn't stick, so trigger a change on the
// abstract compute (the compute all of these radio inputs) are
// bound to so that every radio input can determine whether
// or not it should be "checked".
var abstractCompute = this.options.abstractValue.computeInstance;
can.batch.trigger(abstractCompute, {
	type: 'change',
	triggerOnSameValue: true
});
```

...cascades down to all child events. A better idea might be to configure the specific child computes to trigger change events despite same values. I think I'll look into this. Sounds like the ideal solution.

akagomez:
Okay, I think I have a much better solution in-place for the `type="radio"` part of this PR. 

- Each "radio" compute is bound to the `abstractValue` (as was already the case, only now it has a name)
- Each "radio" compute is configured to trigger `change` events regardless of whether the value has changed or not (by way of `value.computeInstance._triggerOnSameValue`)
- If the `Checked` control attempts to set a "radio" compute's value and fails (meaning the `abstractValue`'s setter rejected/changed our newVal), the `Checked` control manually triggers a `change` event on the `abstractValue`
- The manually triggered `change` event will propagate down to the "radio" computes, thereby causing each `<input type="radio">` to re-evaluate whether or not it is "checked" based on the `abstractValue`'s value. 

As far as I'm concerned this PR is ready for merge, pending review. Please excuse me for the trail of breadcrumbs I left getting to this point.

justinbmeyer:
This is now fixed in 3.0.  Sorry @akagomez for not getting back to you.

PR #1526, opened 2015-03-17
#1521 define map properties in the same order they're defined

### LINKED ISSUE #1521: Using value and get in the same property throws errors
Breaking test showing this behavior here: https://github.com/bitovi/canjs/commit/b45012d612669f801c5211967669e1dc681abaa3

This test passes in https://github.com/bitovi/canjs/releases/tag/v2.2.0-pre.0 but fails in latest master branch.

moschel:
fixes #1521 by enforcing define property ordering

@justinbmeyer and I discussed this afternoon and the root cause was calling the getter for a property that references a default value, but it hadn't been set yet, even though it is before the getter in the property list.

### DIFF
--- a/compute/compute_test.js
+++ b/compute/compute_test.js
@@ -672,10 +672,10 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
 					equal(curVal, 20);
 					break;
 				case 2:
-					equal(curVal, 20, "on bind");
+					equal(curVal, 30, "on bind");
 					break;
 				case 3:
-					equal(curVal, 30, "on bind");
+					equal(curVal, 40, "on bind");
 					break;
 			}
 			return curVal+add.get();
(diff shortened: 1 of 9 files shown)


daffl:
It looks like some tests (AMD, distributables) are failing legitimately. I am also not sure if this is the right fix. The order of keys in objects is not determined and not necessarily the same as it was defined. #1521 is a a regression so it must have been introduced somewhere else.

justinbmeyer:
In practice and in all commonly used versions of JS order is preserved.

moschel:
This is still failing but I'm not sure why, the travis logs don't say which tests failed

justinbmeyer:
It's possible Travis / phantom does not respect order?

daffl:
The reasons for the tests failing is that the attributes plugin does not play very well with define. I made the test conditional since we probably will remove the old attributes plugin - or at least take it out of this repository.

PR #1527, opened 2015-03-17
Update the binding count correctly after calling removeEvent

### LINKED ISSUE #1490: Unbinding non bound property decrements binding counter
This causes the premature object removal from the store as _bindings becomes 0 too early.
See the fiddle: http://jsfiddle.net/donejs/qYdwR/

akagomez:
Closes #1490.

### DIFF
--- a/util/bind/bind.js
+++ b/util/bind/bind.js
@@ -26,13 +26,19 @@ steal('can/util', function (can) {
 		}
 		return this;
 	};
-	can.unbindAndTeardown = function (ev, handler) {
+	can.unbindAndTeardown = function (event, handler) {
+
+		var handlers = this.__bindEvents[event] || [];
+		var handlerCount = handlers.length;
+
 		// Remove the event handler
 		can.removeEvent.apply(this, arguments);
 		if (this._bindings === null) {
 			this._bindings = 0;
 		} else {
-			this._bindings--;
+			// Subtract the difference in the number of handlers bound to this
+			// event before/after removeEvent
+			this._bindings = this._bindings - (handlerCount - handlers.length);
 		}
 		// If there are no longer any bindings and
 		// there is a bindteardown method, call it.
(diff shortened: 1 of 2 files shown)


util/bind/bind.js
```diff
@@ -26,13 +26,19 @@ steal('can/util', function (can) {
...
 		// Remove the event handler
 		can.removeEvent.apply(this, arguments);
 		if (this._bindings === null) {
 			this._bindings = 0;
 		} else {
-			this._bindings--;
+			// Subtract the difference in the number of handlers bound to this
+			// event before/after removeEvent
+			this._bindings = this._bindings - (handlerCount - handlers.length);
 		}
```
justinbmeyer: maybe I'm missing something, but what if someone passes a handler?  Not all events for that handler will be removed.

justinbmeyer:
@daffl @akagomez I think there's a bug here.  But maybe I'm misunderstanding.  What happens if I do:

```
                var map = new can.Map({
			test: 1
		});

		map.on('test', can.noop);
                map.on('test', function(){});
		equal(map._bindings, 2, 'The number of bindings is correct');

                map.off('test', can.noop);
		equal(map._bindings, 1, 'The number of bindings is still correct');
	})
```

akagomez:
@justinbmeyer The test you provided passes: https://github.com/bitovi/canjs/commit/1917e59c4894b54112295062b11c26d07bafbfce

If I remember correctly, `this._bindEvents` is organized like this:

```
{ 
  firstName: [
    {
      name: 'firstName', 
      handler: fn1
    }
  ], 
  lastName: [
    {
      name: 'lastName',
      handler: fn2
    },
    {
      name: 'lastName',
      handler: fn3
    }
  ]
}
```

Let's assume `map.off('lastName', fn3);` is called. 

By getting the `length` of `this._bindEvents.lastName` before and after `can.removeEvent` we know how many event handler were removed for that eventName, and handler/validate method combo. 

Unless I'm wrong in assuming that `this._bindings` is meant to cache the number of eventName/handler pairs there are in `this._bindEvents`, I can't see how this PR introduces a bug.

justinbmeyer:
yeah, you're right.  nice work.

PR #1528, opened 2015-03-17
Adding viewModel to Components control instance

ccummings:
This enables you to use `this.viewModel` inside of a Component like you can use `this.scope` today.

This didn't fail any tests because they are all using the deprecated `this.scope`.


daffl:
Can you change some of the tests so that both are covered? I really thought I caught this one.

PR #1547, opened 2015-03-24
Fixing instances of "recieves" in docs.

Alfredo-Delgado:


Alfredo-Delgado:
Will submit as part of a larger request.

PR #1601, opened 2015-04-03
allow comporator fn return any negative value

whitecolor:
comporator functions should be allowed to return any value less then zero not only -1


daffl:
This makes sense. Is there a way we can test it?

whitecolor:
are the exiting tests passing? from my perspective there should be no problem as we just replace `== -1` with `< 0`

PR #1624, opened 2015-04-16
Adding all my images, HTML, CSS, Less

tomgreever:
Includes all the HTML, CSS, Less, and images for the guides tutorial app. Please move the files wherever you need them for the demo. I just put them in the guides folder.
## CSS

Based on Bootstrap, but only using a few components. Does not include the Bootstrap grid, for example. Less is precompiled into `styles/styles.css` - no need to add a compiler to the project. We could probably remove the Less files altogether, but we'd want to keep them around for maintenance. 
screen shot 2015-04-16 at 10 24 19 am
## Images

All the images came from online services, but I downloaded them into the repo so that the tutorial doesn't require internet access. Each restaurant has a thumbnail, banner, and owner image with sequentially numbered file names. 
screen shot 2015-04-16 at 10 27 59 am
## HTML

All of the markup is in the `index.html` page. Each component is separated by HTML comments like: 
screen shot 2015-04-16 at 10 25 31 am

Screen shots of each component posted here https://github.com/bitovi/canjs/issues/1615#issuecomment-93760237

cc @justinbmeyer @cklanac @imjoshdean @akagomez @joe-crick


akagomez:
@justinbmeyer How do you want to go about the next steps?

PR #1626, opened 2015-04-16
Prevent detached DOM nodes in nested component

### LINKED ISSUE #1625: Nested can.Component causes detached DOM nodes
In this demo (http://jsfiddle.net/akagomez/ajdj2nu2/), when `show` is toggled between `true`/`false`, this markup is added/removed from the DOM:  
```
<parent-component>
  <child-component>
    <ul>
      <li>...</li> // 1000x
    </ul>
  </child-component>
</parent-component>
```

Which creates this memory leak: 

component-leak

Notice how the number of detached DOM nodes is double that of the length of the rendered list.

akagomez:
Fixes #1625. 

I would love to include a test, but I couldn't figure out a way to determine the number of detached DOM nodes/or get a heap snapshot from JS. 

component-leak-fixed

Test used: http://jsfiddle.net/akagomez/ajdj2nu2/

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -274,6 +274,7 @@ steal("can/util", "can/view/callbacks","can/control", "can/observe", "can/view/m
 					can.each(handlers, function (handler, prop) {
 						componentScope.unbind(prop, handlers[prop]);
 					});
+					el = null;
 				});
 
 				// ## `events` control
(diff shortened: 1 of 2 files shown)


daffl:
Can we close this one?

akagomez:
When I spoke with @justinbmeyer yesterday evening, he suggested these were not likely the solution to the core issue. For now, it's likely a good enough fix until Justin has the time to investigate further.

justinbmeyer:
@akagomez thanks for reporting this. It's a good fix for someone who needs a quick fix, but not good enough to merge into master.

imjoshdean:
@justinbmeyer @daffl - is there an ETA on something that would be considered "good enough?" I'd rather not have to fork CanJS to fix such a detrimental bug.

Also this fix only works with Stache - Mustache still has a problem, will that be fixed as well?

daffl:
Closed via #1631

PR #1628, opened 2015-04-16
Prevent detached DOM nodes in nested components that use the <content/> tag

### LINKED ISSUE #1627: Nested can.Component with <content/> tag causes detached DOM nodes
component-leak-2

Code found here: http://jsfiddle.net/akagomez/v8h1qfod/

akagomez:
Fixes #1627.

component-leak-2-fixed

Code being tested in gif: http://jsfiddle.net/akagomez/v8h1qfod/


imjoshdean:
Seems like this should be tied in with #1626, and they're not separate concerns.

PR #1633, opened 2015-04-21
Ignore component accessibilty attrs failing test

cherifGsoul:
The accessibility attributes should be ignored in component viewModel

### DIFF
--- a/component/component_test.js
+++ b/component/component_test.js
@@ -1556,4 +1556,29 @@ steal("can", "can/map/define", "can/component", "can/view/stache" ,"can/route", 
 
 	});
 
+	test('component does not update viewModel on accessiblity attribute changes', function(){
+		
+		can.Component.extend({
+			tag:'x-app'
+		});
+
+		var frag=can.stache('<x-app></x-app>')({});
+		
+		var el = frag.childNodes[0];
+		var viewModel = can.viewModel(el);
+		
+		// element must be inserted, otherwise attributes event will not be fired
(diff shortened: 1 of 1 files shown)


justinbmeyer:
@cherifGsoul Thanks for submitting this.  I think if the viewModel is expecting a role attr, role should be bindable.  We might need a way to make things like id / class / role / etc work.


There's a few problems with the test (I'll comment on those in a second).

component/component_test.js
```diff
@@ -1556,4 +1556,29 @@ steal("can", "can/map/define", "can/component", "can/view/stache" ,"can/route",
...
+		var frag=can.stache('<x-app></x-app>')({});
+		
+		var el = frag.childNodes[0];
+		var viewModel = can.viewModel(el);
+		
+		// element must be inserted, otherwise attributes event will not be fired
+		can.append(can.$("#qunit-fixture"),frag);
+		
+		// update the class
+		can.$(el).attr("role","foo");
```
justinbmeyer: This won't work outside jQuery.  You probably want to use can.attr.set.

cherifGsoul:
The problem is if the viewModel will be in the can.route.map the attribute can be displayed in the url for example: /#!some-path&role=button

cherifGsoul:
@justinbmeyer here is a fiddle http://jsfiddle.net/cherif_b/88drbcxx/ that demonstrate the issue
sorry if im late I was busy with work :)

justinbmeyer:
this is closed for 3.0.

PR #1639, opened 2015-04-22
Add can.fixture.resource method for generating fixtures based on CRUD standards

nickradford:
This method generates fixtures based on standard CRUD operations. Feel free to critique and provide suggestions for a better way to do this.

``` js
var User = can.Model.extend({
  resource: '/users'
},{})

var usersStore = can.fixture.store(100, function(i) {
  return {
    name: 'User ' + i
  }
})

can.fixture.resource('/users', usersStore)
```

/cc @alexisabril

### DIFF
--- a/util/fixture/fixture.js
+++ b/util/fixture/fixture.js
@@ -404,6 +404,37 @@ steal('can/util', 'can/util/string', 'can/util/object', function (can) {
 			});
 			return data;
 		},
+		// ## can.fixture.resource
+		// Create the fixture routes for default CRUD actions
+		resource: function(resource, store, opts) {
+			if (!resource || !store) {
+				return null;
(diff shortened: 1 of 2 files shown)


daffl:
Looks great, thanks! I'll put it in for 2.3. We might have to merge it into the minor branch then but that shouldn't be a problem.

nickradford:
Sounds great @daffl. Do you have any guidance around documentation so I can add some for this feature?

daffl:
I would probably add it to https://github.com/bitovi/canjs/blob/master/util/fixture/doc/store.md. You can also create a similar new page in that folder called `resource`.

justinbmeyer:
can-fixture does this now when passed a store.

PR #1640, opened 2015-04-23
Document setting context of Stache data helper

### LINKED ISSUE #1619: Changing context in a Stache template breaks the data helper
In the migration from EJS to Mustache, we lost the ability to explicitly declare which data was used with the data helper. For example, in EJS you could use:

```<div (el) -> can.data(el, 'todo', todo)></div>```

In some cases, we rewrote the JS to expect the data to be the entire context, but in others, we worked around this limitation by changing the context in the template surrounding the usage of the data helper. For example:

```<div {{#todo}} {{ data "todo" }} {{/todo}}></div>```

Unfortunately, while upgrading to Stache, this broke due to an apparent bug that seems to put JS code into the DOM as exhibited by this fiddle.

http://jsfiddle.net/qYdwR/2476/

So far, this seems to only be a problem with ```{{#key}}{/key}}``` and not with ```{{#each key}}{{/each}}``` which also changes context. A similar issue also occurs with using ```#if``` to wrap around the data helper instead.

marshallswain:
As mentioned in #1619, this documents setting the context of the data helper explicitly.

Does the "optional argument" syntax work in that @signature?

Closes #1619

### DIFF
--- a/view/stache/doc/helpers/data.md
+++ b/view/stache/doc/helpers/data.md
@@ -1,6 +1,6 @@
 @function can.stache.helpers.data {{data name}}
 @parent can.stache.htags 7
-@signature `{{data name}}`
+@signature `{{data name[ context]}}`
 
 Adds the current [can.stache.context context] to the
 element's [can.data].
@@ -12,7 +12,7 @@ context.
 
 ## Use
 
-It is common for you to want some data in the template to be available
+It is common to want some data in the template to be available
 on an element.  `{{data name}}` allows you to save the
 context so it can later be retrieved by [can.data] or
 `$.fn.data`. For example,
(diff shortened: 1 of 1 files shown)


daffl:
Isn't `context` here more the data that are being set?

marshallswain:
I might misunderstand how it works.  I thought the optional argument was to allow setting another key as an alternative to the current scope.  Maybe it should just say `key` instead of `context`?

daffl:
Well I mean basically what you do with `<div {{data 'person' person2}}>` is `$(div).data('person', person2)` so it's simply just the value that you are setting. I find the word `context` potentially confusing here.

alexisabril:
This was fixed by commit b7c8769b26c4b56e866f5d5804c353693bd55ad3

Issue #1643, opened 2015-04-24
Simple syntax for descending sort


nickradford:
It would be nice to have a syntax for supporting descending sort. Writing a manual comparator function is ok, but there could be some sugar that can could provide. Any thoughts on the below code?

```js
var list = new can.List([{
  priority: 4,
  name: 'low'
}, {
  priority: 1,
  name: 'high'
}, {
  priority: 2,
  name: 'middle'
}, {
  priority: 3,
  name: 'mid'
}]);

// Option 1: Adding a predicate before the property.
list.attr('comparator', '-priority');

// Option 2: Using an options object.
list.attr('comparator', { property: 'priority', direction: 'desc'})
```

\cc: @robr24, @alexisabril

justinbmeyer:
cc @akagomez

justinbmeyer:
Closing for: https://github.com/canjs/can-list-sort/issues/2

PR #1644, opened 2015-04-24
Support for comparator objects in list/plugin/sort

nickradford:
This adds support for list sort comparator objects, so sort direction can be provided as opposed to defining a sort function. 

Potential solution for #1643.

``` js
var list = new can.List([{
  priority: 4,
  name: 'low'
}, {
  priority: 1,
  name: 'high'
}, {
  priority: 2,
  name: 'middle'
}, {
  priority: 3,
  name: 'mid'
}, {
  priority: 5,
  name: 'lowest'
}]);

list.attr('comparator', { property: 'priority', direction: 'desc' });
```

TODO:
- [x] Fix whitespace to original format
- [ ] Add documentation

### DIFF
--- a/list/sort/sort.js
+++ b/list/sort/sort.js
@@ -87,6 +87,15 @@ steal('can/util', 'can/list', function () {
 			if (comparator && typeof comparator === 'function') {
 				return comparator(a, b);
 			}
+			
+			if (typeof comparator === 'object' && comparator.property) {
+				a = a[comparator.property];
+				b = b[comparator.property];
+				
+				if (comparator.direction === 'desc') {
+					return a === b ? 0 : a > b ? -1 : 1;
+				}
+			}
 
 			return a === b ? 0 : a < b ? -1 : 1;
 		},
--- a/list/sort/sort_test.js
+++ b/list/sort/sort_test.js
@@ -114,6 +114,72 @@ steal("can/list/sort", "can/test", "can/view/mustache", "can/view/stache", "can/
 		equal(list[0].name, 'high');
 	});
 
+	test('Defining a comparator object with descending direction', 1, function () {
+		var list = new can.List([{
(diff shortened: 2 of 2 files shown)


akagomez:
This idea is starting to grow on me.

nickradford:
I mean, writing a sort function isn't that difficult, but it seemed very strange to me that you could pass in the property name to sort ascending, but to sort the same property descending, you had to pass in a function.

marshallswain:
I liked the shorthand syntax that I saw earlier of 
```
list.attr('comparator', '-property')
```
I don't think it is too unorthodox to be considered.

nickradford:
@marshallswain I agree, but I'm coming from an Angular background where that syntax is familiar. I'm totally open to the idea, but syntaxes are sometimes hard to agree upon.

whitecolor:
are you going to merge this?

daffl:
@gomez Since you know the sort plugin best and if you agree with this PR, feel free to merge it in.

Gomez:
ping @akagomez as i am an other gomez :)

akagomez:
@daffl Thanks. 

@m-mujica and I discussed a few ideas a while back: https://gist.github.com/akagomez/c7d7111e4490e9dd49eb

This one fits right in with the rest besides for the name of the property. If we can change it to `attr` instead of `property` I'll be happy to merge it.

nickradford:
@akagomez I can change `property` to `attr` later today.

julia-allyce:
@nickradford are you still going to make this change? :smile:

akagomez:
I just realized that `direction` should also be renamed to `order`.

gsmeets:
`direction` is the more commonly used term as far as I know.
Models might have an `order` property, which can be sorted in ascending or descending `direction`.

PR #1645, opened 2015-04-24
registerSimpleHelper

daffl:
This pull request adds `registerSimpleHelper` functionality to Mustache and Stache including tests and documentation.

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -255,6 +255,11 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/control", "ca
 						hookupOptions.scope.add(this.scope),
 					options = {
 						helpers: {}
+					},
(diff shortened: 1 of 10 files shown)


daffl:
Minor

imjoshdean:
What's the benefit of a simple handler over a regular handler?

matthewp:
Simple helpers you get the values and not computes. If you've written code like this:

```js
helpers: {
  addOne: function(val) {
    val = val.isComputed ? val() : val;
    return val + 1;
  }
}
```

Then you'd probably like a simpleHelper.

imjoshdean:
Oh, neat! So basically quick and dirty helpers that have zero liability of live binding?

Issue #1651, opened 2015-04-28
Helpers bind to nested properties differently in stache/mustache


akagomez:
Test: https://github.com/bitovi/canjs/commit/68d438b1c62b8491baf95b8b7e1bc6cde30fe58b

kylegifford:
To clarify the problem a bit:

Binding via dot syntax (`this.attr('parent.child')`):
- works with `this.attr('parent.child', 'foo')`
- does not work with `this.attr('parent', {'child': 'foo'})`

Binding via nested attrs (`this.attr('parent').attr('child')`):
- works with `this.attr('parent', {'child': 'foo'})`
- does not work with `this.attr('parent.child', 'foo')`

akagomez:
@kylegifford I think my fix here in #1655 will at least make the two tests consistent.

akagomez:
I don't think binding to...

```
var map = new can.Map({
  parent: null
})
```

With... 

```
map.attr('parent.child')
```

And expecting that to work after.... 

```
map.attr({ parent: { child: 'foo' }}); 
```

Is officially supported. Although, I certainly wish it was. Either way I think it's a separate issue.

kylegifford:
After applying the fix in #1655 the tests work consistently between Stache and Mustache, and binding via nested attrs now works correctly in Stache. Both still fail using dot syntax mentioned above.

@akagomez can this be closed (after cleaning up tests) since this is working consistently in both Stache and Mustache?

akagomez:
That makes sense to me. I'll create a new issue for the dot syntax binding.

kylegifford:
Tests added to #1655 .

daffl:
Closed via #1655

PR #1655, opened 2015-04-30
Don't use setupSingleBindComputeHandlers for Stache helpers

### LINKED ISSUE #1652: can.stache helpers that rely on array replacement don't work, but work in can.mustache
Found this issue earlier today. In a component, if you have a mustache helper that relies on running on a list's length changing, and you replace the list outside of the component - everything works just fine.

If you do this using Stache...the helper stops running.

Example (can.mustache): http://jsfiddle.net/imjoshdean/ojrhg955/
Example (can.stache): http://jsfiddle.net/imjoshdean/ojrhg955/1/

akagomez:
Closes #1652, and #1651.

### DIFF
--- a/view/mustache/mustache_test.js
+++ b/view/mustache/mustache_test.js
@@ -4003,4 +4003,38 @@ steal("can/model", "can/view/mustache", "can/test", "can/view/mustache/spec/spec
 		state.attr('list').push('...')
 
 	});
+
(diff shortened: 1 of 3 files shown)


daffl:
Can you rebase this one, too? Probably just the Stache tests that need to be merged.

kylegifford:
Rebased, should be ready to merge now.

PR #1656, opened 2015-04-30
Add workaround for iOS 8 64-bit JIT object length bug

### LINKED ISSUE #1654: Check for vulnerability against iOS 8 object length bug
[This WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=142792) has exposed issues in [jQuery](https://github.com/jquery/jquery/issues/2145) and [Underscore](https://github.com/jashkenas/underscore/issues/2081). It might be an issue with CanJS because our `isArrayLike` is [from jQuery](https://github.com/bitovi/canjs/blob/cc82e948e2040bdacb6bf1301030eb68de3f86c9/util/array/each.js#L4).

chasenlehara:
I tested this on an iPhone 6 (NG4R2LL/A) with iOS 8.3 (12F70) in Safari by loading the page with the test and seeing it fail without the fix, pass with the fix.

I added a new test file and I’m not completely sure I added it everywhere it needs to be referenced, so please let me know if that or anything else needs to be fixed.

Closes #1654


daffl:
Looks good to me, however, files like `test/dojo.html` etc. are auto generated. It looks like you modified them manually which will probably overwrite the changes. The `util/array/each` module is included everywhere now right? I think we are testing the util/test module so maybe we can just add it there?

chasenlehara:
Does this look correct @daffl? I’m not sure whether “the util/array/each module is included everywhere,” how could I check that?

Sorry for the n00b questions. :)

PR #1658, opened 2015-04-30
Allows can-import to be used for dynamic imports

matthewp:
This adds the ability of can-import to be used for dynamic imports. The
static form is still supported, but dynamic imports are allowed by
including a closing `</can-import>` tag and nesting content within.

The viewModel for `can-import` is a Promise that will resolve with a
`value` once the module has been imported.


matthewp:
Going to re-open against minor.

PR #1659, opened 2015-04-30
Allows can-import to be used for dynamic imports

matthewp:
This adds the ability of can-import to be used for dynamic imports. The
static form is still supported, but dynamic imports are allowed by
including a closing `</can-import>` tag and nesting content within.

The viewModel for `can-import` is a Promise that will resolve with a
`value` once the module has been imported.

### DIFF
--- a/test/jquery-2.html
+++ b/test/jquery-2.html
@@ -56,6 +56,7 @@
                 "can/util/string/string_test.js",
                 "can/util/attr/attr_test.js",
                 "can/view/autorender/autorender_test.js",
+                "can/view/import/import_test.js",
                 function() {
                     can.dev.logLevel = 3;
                 });
--- a/view/import/import.js
+++ b/view/import/import.js
@@ -1,0 +1,20 @@
+steal("can/util", "can/view/callbacks", function(can){
+
+	can.view.tag("can-import", function(el, tagData){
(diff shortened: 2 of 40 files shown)


view/import/import.js
```diff
@@ -0,0 +1,20 @@
+steal("can/util", "can/view/callbacks", function(can){
+
+	can.view.tag("can-import", function(el, tagData){
+		var moduleName = el.getAttribute("from");
+		var importPromise;
+		if(moduleName) {
+			importPromise = window.promise = can["import"](moduleName);
```
matthewp: oops, need to remove this.

test/jquery-2.html
```diff
@@ -56,6 +56,7 @@
                 "can/util/string/string_test.js",
                 "can/util/attr/attr_test.js",
                 "can/view/autorender/autorender_test.js",
+                "can/view/import/import_test.js"
```
daffl: `<!-- AUTO GENERATED - DO NOT MODIFY -->` ?

daffl: I guess I can fix that.

daffl:
Oh also, tests can't be written in CommonJS format yet (but probably should eventually although I'd prefer ES6).

PR #1663, opened 2015-05-01
Make findAll reject on non-array data property.

### LINKED ISSUE #1662: can.Model.findAll should reject if parseModels returns a non-array `data` property.
If you have the following code:

```javascript
can.fixture("/mymodels", function () {
  return {
    status: 'success',
    message: ''
  };
});

var MyModel = can.Model.extend({
  findAll: "/mymodels",

  parseModels: function(raw) {
    raw.data = undefined;
    return raw;
  }
}, {});

MyModel.findAll({}).then(success, error);
```

The `error` callback should be called and not the `success` callback.

m-mujica:
Closes #1662.

### DIFF
--- a/model/model.js
+++ b/model/model.js
@@ -173,7 +173,7 @@ steal('can/util', 'can/map', 'can/list', function (can) {
 					raw = raw.data;
 				}
 
-				if (typeof raw === 'undefined') {
+				if (typeof raw === 'undefined' || !can.isArray(raw)) {
 					throw new Error('Could not get any raw data while converting using .models');
 				}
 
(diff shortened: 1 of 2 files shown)


m-mujica:
The failing test before my fix. 

screen shot 2015-05-01 at 17 07 18

daffl:
Following up on this, I'm wondering why this doesn't already reject since the default `parseModels` does throw an error at https://github.com/bitovi/canjs/blob/parse-models-issue/model/model.js#L261. Are you using your own parseModels?

m-mujica:
yes @daffl I have a custom `parseModels`, you can see it in the test I created https://github.com/bitovi/canjs/pull/1663/files#diff-f5751f58f80284ca23c4942749b780feR1450.

daffl:
This is probably still good to add but your `parseModels` should probably make sure that it is only returning and array and otherwise throw an error itself.

PR #1664, opened 2015-05-01
Fixes issue with setter not setting value if setVal not passed

### LINKED ISSUE #1648: Cannot read value of property defined with "get" AND "set" methods
Demo:
http://jsfiddle.net/akagomez/deagovbn/

Problem 1:
Calling `map.attr('foo')` does not return the correct value.

Problem 2:
`foo`'s `get` method is passed the wrong `lastSetVal` after being `map.attr('foo')` is called. 

Problem 3:
When `foo` is bound to, the `get` method doesn't get called at all. 

I created a test that demonstrates the issue here: https://github.com/bitovi/canjs/commit/369ff13614b0463a8596666bee96135c20bf60d0

/cc @daffl @justinbmeyer

kylegifford:
Fixes #1648.

These tests identified another issue, where assertions in the getter would be called different numbers of times depending on whether the `can.map.attributes` module was included or not. Tests have been refactored to remove these assertions, and a new issue has been opened to address this issue.

### DIFF
--- a/map/define/define.js
+++ b/map/define/define.js
@@ -155,7 +155,7 @@ steal('can/util', 'can/observe', function (can) {
 			if (getter) {
 				// if there's a getter we don't call old set
 				// instead we call the getter's compute with the new value
-				if(setValue !== undefined && !setterCalled && setter.length >= 2) {
+				if(setValue !== undefined && !setterCalled && setter.length >= 1) {
 					this[prop](setValue);
 				}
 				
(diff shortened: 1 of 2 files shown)


akagomez:
The new issue related to `can.map.attributes`: #1665.

daffl:
@akagomez What new issue? I'd argue that can.map.attributes and the define plugin shouldn't be used together (which is I think exactly what the deprecation warning says).

akagomez:
@daffl I only meant to link the issue that @kylegifford referenced in his comment.

PR #1667, opened 2015-05-02
Compute ordering issue in stache and memory leak fix

### LINKED ISSUE #1617: Compute code gets inserted into DOM with Stache sometimes
When you render a Stache template with a binding to a key that is not a can.compute, if it changes to a can.compute, the compute's code will be inserted into the DOM. This does not happen in Mustache.

Here's a fiddle.

http://jsfiddle.net/qYdwR/2471/

moschel:
This fixes #1617 and #1627 by introducing an improved [can.view.Scope.prototype.computeData function](https://github.com/bitovi/canjs/blob/1617-stache-computes/view/scope/compute_data.js) that is able to detect when the "single bind" strategy is no longer valid and switch to a "diff bind" strategy. This is actually slightly faster than the old "single bind", with the added benefit of the "diff bind" working if needed.

@moschel 's original comment:

> fixes #1617
> 
> In the case of the test, this handler is called back with newVal set to a compute function, so its  appropriate to call the function to get the value before triggering the update.
> Without this change, the test outputs the text of a function into the DOM, but the same thing works fine in mustache.


moschel:
the tests work in jQuery, but fail in mootools, YUI, and dojo

moschel:
@daffl any plans to drop those?

moschel:
tests all passed locally, so maybe this is a fluke or a phantom only thing

view/stache/stache_test.js
```diff
@@ -3930,8 +3930,21 @@ steal("can/view/stache", "can/view", "can/test","can/view/mustache/spec/specs","
 
 		equal(frag.childNodes.length, 1, "only the placeholder textnode");
 	});
+	
+	test("#1617 compute defined after template", function(){
```
justinbmeyer: We always put the test number at the end of the test name in parenthesis.

view/stache/stache_test.js
```diff
@@ -3930,8 +3930,21 @@ steal("can/view/stache", "can/view", "can/test","can/view/mustache/spec/specs","
...
+		var myMap = new can.Map();
+
+		// 1. Render a stache template with a binding to a key that is not a can.compute
+		var dom = can.stache('{{ myMap.test }}')({myMap: myMap});
+		can.$('#qunit-fixture').html(dom);
```
justinbmeyer: no reason to add stuff to the DOM unless you have to.  There's no reason here.

PR #1669, opened 2015-05-02
on() doesn't remove non templated event handlers

### LINKED ISSUE #1623: Event handlers attached on Component's init (event) are not called.
The issue is replicated here http://jsbin.com/fucabiyade/1/edit?js,console,output

Basically if you do something like:

```javascript
can.Component.extend({
  tag: 'app-foo',
  events: {
    init: function(){
      this.on(window, 'click', 'doSomething');
    },
    doSomething: function(){
    }
  }
})
```

`doSomething` won't be called if you click around, it does work if the event handler is attached on the `inserted` event, or if you do (a more idiomatic way I think) `{window} click: 'doSomething'`. This used to work before CanJS 2.2, probably a regression?

moschel:
fixes #1623 

Control's on method was calling off, which was removing all event handlers, including those that were added by calling this.on('click', fn) and not just those that matched patterns and therefore had 'actions'.

This meant that any time a user would add an event handler like this.on('click', fn) and then call this.on() those event handlers were being wiped out. This fixes that problem with some additional logic in off.


component/component_test.js
```diff
@@ -1558,6 +1558,31 @@ steal("can", "can/map/define", "can/component", "can/view/stache" ,"can/route",
...
+				doSomethingfromInserted: function(){
+					ok(true, "bound in inserted");
+				},
+				doSomethingfromInit: function(){
+					ok(true, "bound in init");
+				}
+			}
+		});
+		can.append( can.$("#qunit-fixture"), can.stache("<app-foo></app-foo>")({}));
+		can.$("#qunit-fixture div").trigger('click');
```
daffl: This needs to be [can.trigger(el, ev)](http://canjs.com/docs/can.trigger.html) then tests should pass with the other libraries as well.

moschel:
@daffl what's holding this one up?

daffl:
The tests are still failing (I'll have to look at it again) and @m-mujica reported this as a regression in 2.2 which doesn't make sense because this code hasn't really changed in about a year. I'd like to find out why this broke something that used to work before first.

daffl:
This fails some backwards compatibility tests as well (see https://travis-ci.org/bitovi/canjs/jobs/62603307#L2571).

justinbmeyer:
> This meant that any time a user would add an event handler like this.on('click', fn) and then call this.on() those event handlers were being wiped out.

This has always been true right?

daffl:
I created a better fix in #1803.

PR #1671, opened 2015-05-03
adds a deprecation warning for can.mustache

### LINKED ISSUE #1638: Mark `can.mustache` as deprecated & add note to `can.Component`
With the memory leak fixes that were made to `can.Component` as they relate to `can.stache` exclusively - we have officially turned out backs on `can.mustache`. We should put a deprecated tag on it in the API and perhaps a note on `can.Component` urging people that if they intend to use components then they should use `can.stache` instead, otherwise there is the potential for bad memory leaks.

moschel:
fixes #1638 

screenshot 2015-05-02 22 15 01

### DIFF
--- a/component/component.md
+++ b/component/component.md
@@ -157,6 +157,8 @@ Changes `<hello-world>Hi There</hello-world>` into:
 
     <hello-world><h1>Hi There</h1></hello-world>
 
+It is recommended to use [can.stache] with can.Component instead of [can.mustache]. [can.mustache] is deprecated as of 2.2, and [can.stache] will provide improved performance and memory usage.
+
 ### viewModel
 
 A component's [can.Component::viewModel viewModel] defines a can.Map that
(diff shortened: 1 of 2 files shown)


daffl:
@James0x57 wrote a migration script for this. Before putting this warning in we should polish it, add documentation and link to it.

moschel:
@James0x57 is this script done? Can you point to where it is?

James0x57:
Hey Guys, here is the script:
https://github.com/canjs/can-upgrade/blob/master/bin/mustache_to_stache.js

There is one TODO note in there I'd like to address when I have time, and I should probably remove the string prototype getters I did for colors and do them another way, but it all works as is.

$ cd /root/of/your/project
$ npm install can-upgrade
$ node /node_modules/can-upgrade/bin/mustache_to_stache.js .

daffl:
It would be great if you could add some quick documentation in the Readme so that we can link it in the deprecation warning.

julia-allyce:
@James0x57 are we ok to link to can-upgrade yet? If not, I think we should merge this PR so there is at least a notice in for mustache.

James0x57:
@julia-allyce should be fine to link to

pYr0x:
i think we can close this.
there is a deprecated warning and a link to can-upgrade

PR #1674, opened 2015-05-04
Update instructions for steal

marknadig:
Direct user to Can2.2/Steal blog post for current info.


daffl:
Good call. Once I finished the other articles I will try and incorporate them into the guides but linking them for now makes sense.

PR #1678, opened 2015-05-08
Fixing some mistakes

dylanrtt:
The ```<my-widget>``` component is never defined and seems to be referring to the ```<hello-world>``` component.


daffl:
I think this one is conflicting with #1661 I just merged. If you rebase with latest master I will merge this PR right away. Thanks!

Issue #1676, opened 2015-05-07
Compute memory leak with nested properties


dylanrtt:
Apparently if you manually add a compute to a Map, and the compute depends on a deep property of that same Map, it can leak bindings if used in conjunction with other event bindings in a certain order.

http://jsbin.com/nadelesulu/1/edit?html,js,console

justinbmeyer:
This is likely due to can.Map thinking that compute represents a define.get.

justinbmeyer:
Looked into this.  First, I don't think you should be writing your code like this.  There's no reason to bind to "change" events on `can.Map` anymore and you could use the define plugin to describe `petInfo`.

I'm not sure why you would set a compute as a property value of an attribute either.


All that being said, it would still be nice to make this work.  The problem is that this code creates a "cycle" of binding that isn't being detected while unbinding.

justinbmeyer:
#1231 is probably related

justinbmeyer:
Closed by #1696.

Issue #1657, opened 2015-04-30
Cannot bind to future nested properties


akagomez:
```
var state = new can.Map({
	parent: null
});

var helpers = {
	childExists: function (options) {
		return this.attr('parent.child') ?
			options.fn() :
			options.inverse();
	}
}

var frag = can.stache('{{#childExists}}{{/childExists}}')(state, helpers);

// Does not cause helper to be re-evaluated
state.attr('parent', {
	child: 'foo'
}); 
```

I'm not sure if this is meant to be supported, but it sure would be awesome not to have to do `this.attr('parent') && this.attr('parent.child')` for situations like these. 

Spun off from #1651. 

/cc @kylegifford

justinbmeyer:
This is due to where can.__reading is called within can.Map.  I tried to move it into _get, where I think it belongs, but a bunch of things broke.  I don't remember why, but that is the solution.

akagomez:
Awesome! I'll run with that.

justinbmeyer:
related to #1231

justinbmeyer:
Closed by #1696.

Issue #1231, opened 2014-09-15
can.compute nested key behaviour inconsistency with map bindings


stevenvachon:
`can.compute`s with `attr("value.nested")` do not behave the same as `attr("value").attr("nested")` in a map binding.

http://jsfiddle.net/prometh/La806u2d/

chasenlehara:
+1, we ran into [an issue caused by this bug in Sam’s](https://github.com/bitovi/samsclub/issues/4218).

justinbmeyer:
Closed by #1696.

PR #1696, opened 2015-05-14
Cleans up computes and fixes deep .attr calls.

justinbmeyer:
I pull out most of the "computed" behavior into a compute/observe.js file:  https://github.com/bitovi/canjs/commit/d2eabccc736ba1425d5744d526ab56fa3231cc8c#diff-2ca5aa1b3dd10b09d026b426e2ebf950R1

Did a little inline docs too.


daffl:
What are the related issues? #1231, #1676 and #1657 right?

justinbmeyer:
@daffl yes.

PR #1702, opened 2015-05-15
Use /latest jsbins

marshallswain:
This makes it possible to modify the jsbin without having to make another commit.  You can lock bins in the jsbin UI to prevent changes once you have everything the way you want it.


daffl:
We should look into migrating more JSBins.

Issue #1700, opened 2015-05-15
Figure out attribute binding syntaxes


justinbmeyer:
In combination with #1699, this proposal seeks to create a more flexible, powerful, and easily understood attribute binding syntaxes for use in stache.

## Summary

| Type                 | Example  |
|-------------------------|--------------------------|
| event - viewModel | `<my-comp (child-event)="parentMethod()"/>` |
| event - DOM | `<my-comp ($click)="parentMethod()"/>` |
| 0 way - viewModel  | `<my-comp child-prop="foo"/>` |
| one way - viewModel - parent to child | `<my-comp {child-prop}="parentProp"/>` |
| one way - DOM - parent to element | `<my-comp {$el-prop}="parentProp"/>` |
| one way - reference - child to reference | `<my-comp {^child-prop}="*ref"/>`        |
| two way - viewModel | `<my-comp {(child-prop)}="parentProp"/>`  |
| two way - DOM | `<my-comp {($child-prop)}="parentProp"/>`  |
| two way - reference | `<my-comp {(child-prop)}="*ref"/>`        |
| two way - reference shorthand | `<my-comp *ref/>`        |

## event - viewModel

`(event-name)="StacheMethodExpression"`

Binds an event on the `can.viewModel` of whatever element it was placed on.

Example:

```
<modal (todo.created)="hideModal()" (cancelled)="hideModal()"/>
```

## event - DOM

`($event-name)="StacheMethodExpression"`

Bind an event on the element it was placed on.

Example:

```
<button ($click)="todo.save()">Save</button>
```

## 0 way - viewModel

`view-model-attr="value"`

Sets the `viewModelAttr` property to `"value"`.

## one way - viewModel - parent to child

The following looks up `task` in the scope, and sets it as `todo` in the `<modal>`'s viewModel:

```
<modal {todo}="task"/>
```

## one way - DOM - parent to element

Provides one-way binding to an element's property or attribute.  We will maintain a list of properties that will be set instead of setting the attribute.

```
<input {$checked}="complete" type='checkbox'/>
```

This is equivalent to:

 ```
<input {{#complete}}checked{{/complete}} type='checkbox'/>
```

While it's possible to do one-way binding to an element property or attribute with stache alone, two-way binding uses this for syntactic sugar. 

## one way - reference - child to reference

The following looks up `task` in the child viewModel, and sets it as `todo` in the current template's 
reference scope:

```
<modal {^task}="*todo"/>
```

## two way - viewModel

[...]

PR #1710, opened 2015-05-23
Reference values 

justinbmeyer:
For #1700, this adds reference values exporting like:

```
<my-component #foo>
```

This also cherry-picks the parent context exporting:

```
<my-component [parentProp]="childProp"/>
```

- [ ] -This should be adding `#[\w\.]+` to `var ignoreAttributesRegExp`, but it is not.  
- [ ] - This needs docs.

This uses a new `can.view.Scope` method, `cloneFromRef` to make a context for "light dom" that uses the lexical semantics for the "refs".  `cloneFromRef` essentially copies a Scope, but with the component's refs removed.

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -255,6 +255,11 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/control", "ca
 						hookupOptions.scope.add(this.scope),
 					options = {
 						helpers: {}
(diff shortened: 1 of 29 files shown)


justinbmeyer:
Wanted to merg this into minor

PR #1717, opened 2015-05-26
Virtual DOM and place-my-order functionality

daffl:
This branch contains the virtual DOM implementation and other functionality for place-my-order.


matthewp:
I updated can-simple-dom so that we don't have to do `simpleDOM["default"]` any more.  All steal tests are passing, most of the dist tests are passing, nothing is failing that was passing before.  I'm not comfortable pushing to a branch that's not passing so if I broke something it's not because i didn't test.

matthewp:
Green!

daffl:
Finally! That was a beast ;)

Issue #1707, opened 2015-05-22
Sort plugin excludes batchNum; Maybe fires unnecessary events


akagomez:
```
can.batch.start();
listB.push({ id: 'a' });
listB.push({ id: 'a' });
listB.push({ id: 'a' });
can.batch.stop();
```

```
"// BATCHED; SORTED..."
"    Sorted..." // <-- Upon creating the list
"    Sorted..." 
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the sort
"    Swapped? 1, 2"
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the swap
["type:", "length", "batchNum:", 9]
```

The Sort plugin fires events like `length`, `add`, `reset`, etc. for `can.view.live.list` and does nothing to pass along the original `batchNum` that instigated the work. 

Since the current Sort plugin is derived from the previously incomplete Sort plugin, those events were left in/adapted. I'm curious to know if we can do without them all together. 

http://jsbin.com/poduxa/6/edit?js,console

daffl:
Closed via #1718

Issue #1706, opened 2015-05-22
Sort plugin evaluates insert index for batched events


akagomez:
```
can.batch.start();
listB.push({ id: 'a' });
listB.push({ id: 'a' });
listB.push({ id: 'a' });
can.batch.stop();
```

```
"// BATCHED; SORTED..."
"    Sorted..." // <-- Upon creating the list
"    Sorted..." 
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the sort
"    Swapped? 1, 2"
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the swap
["type:", "length", "batchNum:", 9]
```

The first two `sort`'s make sense. One for calling `new` the, next for having batched changes. The `swap` happens as a result of `_change` being called subsequently with the same `batchNum`.

http://jsbin.com/poduxa/6/edit?js,console

daffl:
Closed via #1718

PR #1718, opened 2015-05-27
Handle a few sort plugin edge cases

### LINKED ISSUE #1705: Sort plugin swaps equal value items
```
can.batch.start();
listB.push({ id: 'a' });
listB.push({ id: 'a' });
listB.push({ id: 'a' });
can.batch.stop();
```

```
"// BATCHED; SORTED..."
"    Sorted..." // <-- Upon creating the list
"    Sorted..." 
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the sort
"    Swapped? 1, 2"
["type:", "length", "batchNum:", undefined]
["type:", "length", "batchNum:", 9] // <-- This event caused the swap
["type:", "length", "batchNum:", 9]
```

Since all of the values are the same, no "swap" should occur.

http://jsbin.com/poduxa/6/edit?js,console

akagomez:
- Batched list operations fired events without a `batchNum` (#1707)

  Swapped out all of the `can.trigger`'s with `can.batch.trigger` and made it
  so that `can.batch.trigger` adds the `can.batch.batchNum` when creating an
  `event` object.

- Batched list operations evaluated insert index despite `batchNum` (#1706)

  Fixed the logic that returns/sorts based on the current event's `batchNum`

- `change` events caused list items to be swapped with same-value siblings (#1705)

  The default selection sort is kind of greedy in that it continues to
  search for an insert index until it finds an item with a greater value,
  or the end of the list. This means that it will move an item passed
  siblings of the same value every time the `_getInsertIndex` is evaluated.
  Modifying the algorithm to swap to items that are equal OR greater doesn't
  fix the issue either, it just reverses the direction. The fix was to
  check if the item to the right of its current position is the same value. If it is, 
  don't move it. It's fine where it's at. 

- `change` events evaluated insert index when irrelevant values changed (#1722 fix 1 of 2)

  Previously, `change` events would naively call `_getInsertIndex` even if
  the property that changed wasn't the comparator. This was inefficient since
  it resulted in a loop over the items in the list needlessly. Now the changed
  `attr` is compared to `comparator` on every `change` event. If `comparator`
  is a function OR the changed attr is a substring of the comparator value, the
  insert index is evaluated.

- Added a few tests to make development a bit easier

### DIFF
--- a/list/sort/sort.js
+++ b/list/sort/sort.js
@@ -91,33 +91,42 @@ steal('can/util', 'can/list', function () {
 			return a === b ? 0 : a < b ? -1 : 1;
 		},
 		_changes: function (ev, attr, how, newVal, oldVal) {
+			var dotIndex = ("" + attr).indexOf('.');
 
 			// If a comparator is defined and the change was to a
 			// list item, consider moving the item.
-			if (this.comparator && /^\d+/.test(attr)) {
-	
-				if (ev.batchNum && ev.batchNum !== this._lastBatchNum) {
-					this.sort();
-					this._lastBatchNum = ev.batchNum;
-					return;
+			if (this.comparator && dotIndex !== -1) {
+				if (ev.batchNum) {
+					if (ev.batchNum === this._lastProcessedBatchNum) {
+						return;
+					} else {
+						this.sort();
+						this._lastProcessedBatchNum = ev.batchNum;
+						return;
+					}
 				}
 	
-				// get the index
-				var currentIndex = +/^\d+/.exec(attr)[0],
-					// and item
-					item = this[currentIndex];
-	
-				if (typeof item !== 'undefined') {
+				var currentIndex = +attr.substr(0, dotIndex);
+				var item = this[currentIndex];
+				var changedAttr = attr.substr(dotIndex + 1);
+
+				// Don't waste time evaluating items in ways that aren't
+				// relevant or have changed in ways that aren't relevant.
+				if (typeof item !== 'undefined' &&
+					(typeof this.comparator !== 'string' ||
+						this.comparator.indexOf(changedAttr) === 0)) {
 	
 					// Determine where this item should reside as a result
 					// of the change
-					var newIndex = this._getInsertIndex(item, currentIndex);
-	
+					var newIndex =
+						this._getRelativeInsertIndex(item, currentIndex);
+
 					if (newIndex !== currentIndex) {
 						this._swapItems(currentIndex, newIndex);
 	
-						// Trigger length change so that {{#block}} helper can re-render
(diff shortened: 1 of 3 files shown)


util/batch/batch.js
```diff
@@ -216,7 +216,8 @@ steal('can/util/can.js', function (can) {
 			// Don't send events if initalizing.
 			if (!item._init) {
 				event = typeof event === 'string' ? {
-					type: event
+					type: event,
+					batchNum: can.batch.batchNum
```
akagomez: @justinbmeyer Am I right to assume that this is a reasonable default? It seemed correct based on how I understand `can.batch.batchNum` to work. I wondered if it should be added to the other `event` objects created [here](https://github.com/bitovi/canjs/blob/master/util/jquery/jquery.js#L16-L20) and [here](https://github.com/bitovi/canjs/blob/master/event/event.js#L241-L245) also.

list/sort/sort_test.js
```diff
@@ -522,4 +522,130 @@ steal("can/list/sort", "can/test", "can/view/mustache", "can/view/stache", "can/
...
+
+		// Start test
+		list.attr('0.index', 4);
+		list.attr('comparator', 'child.grandchild.id');
+		list.attr('1.child.grandchild.index', 4);
+
+		list._getRelativeInsertIndex = function () {
+			ok(true, 'This item should be evaluated independently');
+			return _getRelativeInsertIndex.apply(this, arguments);
+		};
```
akagomez: @daffl Is it okay that I'm testing the internal API of the sort plugin this way?

list/sort/sort.js
```diff
@@ -91,33 +91,43 @@ steal('can/util', 'can/list', function () {
...
+			if (this.comparator && dotIndex) {
+
+				if (ev.batchNum) {
+					if (ev.batchNum === this._lastProcessedBatchNum) {
+						return;
+					} else {
+						this.sort();
+						this._lastProcessedBatchNum = ev.batchNum;
+						return;
+					}
```
akagomez: My previous batch handling was joke, which inspired the `can.batch.debounce` discussion here: https://github.com/bitovi/canjs/issues/1708

util/batch/batch.js
akagomez: TODO: Add test

akagomez: Justin agreed this is fine here, but not so in the other two places.

list/sort/sort_test.js
daffl: I would say yes. If we change the internal API we'll have to change the tests which is fine though.

PR #1720, opened 2015-05-28
adds test for and fixes issue with object property computes

### LINKED ISSUE #1719: Provide a getter and setter compute interface for non-map-like targets
```
this._get = can.proxy(this._get, target);
this._set = can.proxy(this._set, target);
```

kylegifford:
Fixes #1719 . Provides a getter and setter compute interface for non-map-like targets.

### DIFF
--- a/compute/proto_compute.js
+++ b/compute/proto_compute.js
@@ -289,8 +289,16 @@ steal('can/util', 'can/util/bind', 'can/compute/read.js','can/compute/get_value_
 					return target.unbind(eventName || propertyName, handler);
 				};
 			} else {
-				this._get = can.proxy(this._get, target);
-				this._set = can.proxy(this._set, target);
+				this._get = function() {
+					return can.getObject(propertyName, [target]);
+				};
+				this._set = function(value) {
+					// allow setting properties n levels deep, if separated with dot syntax
+					var properties = propertyName.split("."),
+						leafPropertyName = properties.pop(),
+						targetProperty = can.getObject(properties.join('.'), [target]);
+					targetProperty[leafPropertyName] = value;
+				};
 			}
 		},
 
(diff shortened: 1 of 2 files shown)


compute/proto_compute.js
```diff
@@ -289,8 +289,12 @@ steal('can/util', 'can/util/bind', 'can/compute/read.js','can/compute/get_value_
 					return target.unbind(eventName || propertyName, handler);
 				};
 			} else {
-				this._get = can.proxy(this._get, target);
-				this._set = can.proxy(this._set, target);
+				this._get = function() {
+					return can.getObject(propertyName, [target]);
+				};
+				this._set = function(val) {
+					target[propertyName] = val;
```
justinbmeyer: This won't work with a property like "foo.bar".

observe/observe_test.js
```diff
@@ -1384,4 +1384,28 @@ steal('can/util', "can/observe", 'can/map', 'can/list', "can/test", "steal-qunit
...
+		}
+		value.bind('change', handler);
+		value('qux');
+		value.unbind('change', handler);
+		stop();
+		setTimeout(function() {
+			obj.foo = 'norf';
+			equal(value(), 'norf', 'property changed correctly out of call stack');
+			start();
+		}, 50);
```
justinbmeyer: is this setTimeout necessary?  Also, it 50ms is a long time when there are 1000 tests.

compute/proto_compute.js
```diff
@@ -289,8 +289,16 @@ steal('can/util', 'can/util/bind', 'can/compute/read.js','can/compute/get_value_
...
 			} else {
-				this._get = can.proxy(this._get, target);
-				this._set = can.proxy(this._set, target);
+				this._get = function() {
+					return can.getObject(propertyName, [target]);
+				};
+				this._set = function(value) {
+					// allow setting properties n levels deep, if separated with dot syntax
+					var properties = propertyName.split("."),
+						leafPropertyName = properties.pop(),
```
daffl: Wouldn't it be faster to use subtr and `properties.lastIndexOf('.')`?

PR #1727, opened 2015-06-02
Fix component partials

ccummings:
This PR allows you to pass renderer functions to components. It will also enable renderer functions in scope to work properly in stache (cherry-picked from this [commit](https://github.com/bitovi/canjs/commit/93890b422f40540bcd0468f0f2b3e10e227a2729) in minor).

JSBin example: http://jsbin.com/picumobuza/1/edit

```js
can.Component.extend({
  tag: 'can-demo',
  template: can.stache('{{> sub}} World'),
  viewModel: {
    define: {
      sub: {
        value: function(){
          return can.stache('Hello');
        }
      }
    }
  }
});

var template = can.view('#template');

can.$('#demo').html(template({
  myTemplate: can.stache('Howdy')
}));
```

```html
<can-demo></can-demo> <!-- Hello World -->
<can-demo sub="{myTemplate}"></can-demo> <!-- Howdy World -->
```

### DIFF
--- a/compute/read.js
+++ b/compute/read.js
@@ -108,7 +108,8 @@ steal("can/util", function(can){
 			return type === 'function' && !value.isComputed &&
 				(options.executeAnonymousFunctions || (options.isArgument && i === reads.length) ) &&
 				!(can.Construct && value.prototype instanceof can.Construct) &&
-				!(can.route && value === can.route);
+				!(can.route && value === can.route) &&
+				!(!!value.render);
 		},
 		read: function(value, i, reads, options, state, prev){
 			if (options.isArgument && i === reads.length) {
--- a/compute/compute_test.js
+++ b/compute/compute_test.js
@@ -332,6 +332,19 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
 		var result = can.compute.read(parent, reads);
 		equal(result.value, Todo, 'Got the same Todo');
 	});
+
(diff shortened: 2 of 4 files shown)


ccummings:
The use case is highly reusable components that allow part of their templates to render a partial. Basically what shadow DOM's content-select would give us.

A concrete example is an autocomplete component where the template that renders each result is passed in.

With 2.3.0 you still wouldn't be able to pass a renderer function to a Component without the can.compute.read fix. i.e I don't think this would work:

```html
<can-import from="other.stache!" #other="{value}"/>
<can-demo my-template="{other}"></can-demo>
```

Even if it was made to work without the can.compute.read fix, I'm using the CommonJS build with Browserify which is incompatible with can-import.

compute/read.js
```diff
@@ -108,7 +108,8 @@ steal("can/util", function(can){
 			return type === 'function' && !value.isComputed &&
 				(options.executeAnonymousFunctions || (options.isArgument && i === reads.length) ) &&
 				!(can.Construct && value.prototype instanceof can.Construct) &&
-				!(can.route && value === can.route);
+				!(can.route && value === can.route) &&
+				!(!!value.render);
```
daffl: `!!!value.render`?

daffl:
Can we close this due to 2.3.0?

daffl:
I'm going to close this. Partials should work correctly in latest.

PR #1771, opened 2015-07-07
#1756 Add can.route.map back into documentation

### LINKED ISSUE #1756: can.route.map is missing from the 2.2 & 2.3 docs
http://canjs.com/2.1/docs/can.route.map.html

lukewhyte:
Closes #1756

### DIFF
--- a/route/route_map.md
+++ b/route/route_map.md
@@ -1,0 +1,153 @@
+@function can.route.map map
+@parent can.route.static
+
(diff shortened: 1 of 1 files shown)


daffl:
Any idea why that files was gone?

lukewhyte:
It appears we lost it in this commit: https://github.com/bitovi/canjs/commit/37111e1c6990c8a4fc244f1da65d6ac502a85876 

Prior to this it was stored in route/docs/ dir, which I believe was removed in the same commit.

PR #1780, opened 2015-07-13
Only attempt unbind when __bindEvents is defined

### LINKED ISSUE #1779: Unbinding from non-bound compute throws error
No error:

```
var c = can.compute(32);
c.bind('change', can.noop);
c.unbind();
```

Error:

```
var c = can.compute(32);
c.unbind(); //-> "TypeError: Cannot read property 'undefined' of undefined
                 at can.unbindAndTeardown [as unbind] 
```

http://jsbin.com/nixatopuye/edit?js,console,output

akagomez:
An error is thrown if `unbind` is called on a compute that has no bindings. 

This PR addresses the issue by returning early when `__bindEvents` is falsy, as is done in  `event.js` here: https://github.com/bitovi/canjs/blob/master/event/event.js#L196-L198

Closes #1779

### DIFF
--- a/compute/compute_test.js
+++ b/compute/compute_test.js
@@ -792,4 +792,10 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
 
 		comp('test');
 	});
+
+	test('Calling .unbind() on un-bound compute does not throw an error', function () {
+		var count =  can.compute(0);
+		count.unbind('change');
+		ok(true, 'No error was thrown');
+	});
 });
--- a/util/bind/bind.js
+++ b/util/bind/bind.js
@@ -27,6 +27,9 @@ steal('can/util', function (can) {
 		return this;
 	};
 	can.unbindAndTeardown = function (event, handler) {
+		if (!this.__bindEvents) {
+			return this;
+		}
 
 		var handlers = this.__bindEvents[event] || [];
 		var handlerCount = handlers.length;


justinbmeyer:
This error can and has served as a good warning that something is wrong. Are we sure we want to remove it?

akagomez:
I can't imagine we'd want to throw that error for the case I outlined. 

What others exist? Can you think of some off hand?

justinbmeyer:
I'd want an error thrown for that case.

justinbmeyer:
Calling unbind without arguments is invalid. Calling unbind that doesn't actually unbind is invalid. 

I've never been a fan of how jQuery swallows errors. I don't think we should do it too.

akagomez:
Maybe there's a better alternative to my use case then. How I can I remove all bindings from a compute?

justinbmeyer:
Maybe is misunderstood what you are trying to do. Removing the error is different then making it possible to not pass an event handler and have all handlers removed.

akagomez:
You're right. I assumed that calling `.unbind()` without any arguments would remove all of the bindings. After getting this error I then assumed that if there were bindings to be removed, they would be. 

I could swear I've confirmed this before. But now I'm not sure. I probably got this mixed up with how `can.Map` can unbind all handlers bound to a property. http://jsbin.com/yomimiyuce/edit?js,console,output

akagomez:
One slight clarification. Calling `.unbind()` without any arguments does in fact not remove any bindings. 

You can remove all of the bindings of a particular type by calling `.unbind([type])`. 

That said, doing so will still throw an error if there are no bindings at the time `.unbind()` is called on the compute. This PR addresses that specific issue. 

Also, test added.

justinbmeyer:
Looks good.  @akagomez anyway you can resolve the merge conflicts, make sure travis passes, and merge this into master?

akagomez:
Rebased. Tests passing. Good to merge.

justinbmeyer:
@akagomez cc @daffl I think I misunderstood what this pull request does.  @akagomez can you confirm that it does not help actually remove all event bindings?  If this is the case, I still don't understand why this is useful.  An error is useful.

akagomez:
The confusion may be due to the fact that I opened the issue for the case `.unbind()`. That was in error. Calling `.unbind()` without an event "type" does nothing. See here: https://github.com/bitovi/canjs/blob/master/event/event.js#L199

I attributed that error to calling `.unbind()` without a type. In reality it was present even if a "type" was defined: 

```
var c = can.compute(32);
c.unbind('change'); //-> "TypeError: Cannot read property 'undefined' of undefined
                 at can.unbindAndTeardown [as unbind] 
```

This PR prevents an error from being be thrown if `.unbind("type")` is called and there are no bindings, which is consistent with most "remove" api's that I'm familiar with (e.g. `el.removeClass('class')`, `map.removeAttr('attr')`, etc)

Making `.unbind()` remove all event handlers of all types was not the intention of this PR.

PR #1782, opened 2015-07-13
Live-list diffing

justinbmeyer:
This adds minimal diffing of lists in can.view.live.list.  This can significantly improve performance if a new list is being generated each time a slight change happens.


matthewp:
@daffl can we go ahead and merge this?

PR #1788, opened 2015-07-17
Use NodeList.prototype.item to get a childNode

matthewp:
Instead of:

```js
var node = parent.childNodes[0];
```

we always need to do:

```js
var node = parent.childNodes.item(0);
```

To support the virtual DOM environment.

### DIFF
--- a/view/stache/html_section.js
+++ b/view/stache/html_section.js
@@ -1,6 +1,6 @@
 steal("can/util","can/view/target","./utils.js","./mustache_core.js",function( can, target, utils, mustacheCore ) {
 
-	
+
 	var decodeHTML = typeof document !== "undefined" && (function(){
 		var el = document.createElement('div');
 		return function(html){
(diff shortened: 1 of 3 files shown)


matthewp:
Found one more.

view/stache/html_section.js
```diff
@@ -8,25 +8,25 @@ steal("can/util","can/view/target","./utils.js","./mustache_core.js",function( c
...
 	// ## HTMLSectionBuilder
 	//
 	// Contains a stack of HTMLSections.
 	// An HTMLSection is created everytime a subsection is found. For example:
-	// 
+	//
 	//     {{#if items}} {{#items}} X
 	//
-	// At the point X was being processed, their would be 2 HTMLSections in the 
+	// At the point X was being processed, their would be 2 HTMLSections in the
```
Alfredo-Delgado: s/their/there

Issue #1415, opened 2015-01-23
removing bound scope properties on destroy


moschel:
By the time the `destroy` event of a component is run, the bound scope properties are no longer bound to their parent component properties. In other words, if you pass a cross bind a scope property from a parent to child component like:

````html
<my-component product="{product}"></my-component>
````

Then you want to clean up that product reference while the component is destroyed like:

````js
events: {
  destroyed: function(){
    this.scope.attr('product', null);
  }
}
````

This won't throw an error, but the product property of the parent component won't be affected by this, because the property bindings are unbound before destroyed is called. I think we should change the order so the bindings are cleaned up after destroyed is called.

moschel:
@daffl can't we close this?

daffl:
No. Referenced issues will be closed automatically once in master.

justinbmeyer:
We can close them before that if they are in minor. Too bad github doesn't have some setting like this.

daffl:
We could make minor the default branch. I don't think that finding all the still open but referenced issues in merged PRs against minor is necessary though. I'd also rather have people watching an issue be notified if a new release has been made not just if the fix made it in a development branch.

PR #1803, opened 2015-07-24
Make sure component init event handlers are not unbound

### LINKED ISSUE #1623: Event handlers attached on Component's init (event) are not called.

daffl:
Closes #1623 which was caused by a regression introduced in #1422 (fix for issue #1415) calling `this._control.on()` removed all event handlers set up during `init`. Now we just pass a function to the Component control that will be called at the right time in `destroy`.

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -282,22 +282,9 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/control", "ca
 				this._control = new this.constructor.Control(el, {
 					// Pass the viewModel to the control so we can listen to it's changes from the controller.
 					scope: this.scope,
-					viewModel: this.scope
+					viewModel: this.scope,
+					destroy: callTeardownFunctions
 				});
-
-				// If control has a 'destroy' event, unbind the properties after its called #1415
-				if(this._control && this._control.destroy){
-					var oldDestroy = this._control.destroy;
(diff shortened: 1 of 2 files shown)


component/component.js
```diff
@@ -478,6 +465,11 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/control", "ca
 			// Call `can.Control.prototype.off` function on this instance to cleanup the bindings.
 			can.Control.prototype.off.apply(this, arguments);
 			this._bindings.readyComputes = {};
+		},
+		destroy: function() {
+			if(typeof this.options.destroy === 'function') {
```
justinbmeyer: Is there a case where this.options.destroy would not be a function?

PR #1815, opened 2015-07-29
Pass options to helpers

### LINKED ISSUE #1783: Parent helpers inside helper
I need to create helper that will return template stache fragment and insert it in DOM, so I do:

```javascript
helpers: {
    someTemplateHelper: function(options){
            var template = '<my-comp"></my-comp>'
            return can.stache(template)(options.contexts)
    }
}
```

The parent context are passed to stache fragment, but not helpers (`options` do not contain `helpers`). Is it possible to access parent helpers and render template that can access them?

matthewp:
This is so helpers can pass the parent helpers to child templates.
Thanks @whitecolor. Fixes #1783

### DIFF
--- a/view/stache/mustache_core.js
+++ b/view/stache/mustache_core.js
@@ -300,7 +300,9 @@ steal("can/util",
 					contexts: scope,
 					hash: hash,
 					nodeList: nodeList,
-					exprData: exprData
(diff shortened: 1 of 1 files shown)


whitecolor:
Thanks.

PR #1820, opened 2015-07-30
Fix typo in default generator dependent property resolution logic

akagomez:
Default value generators that depend on other default value generated properties where not being resolved when the dependent property was read without the dot-syntax. 

In other words, this worked:

```
define: {
  user: {
    firstName: 'Chris', 
    lastName: 'Gomez'
  },
  userFullName: {
    value: function () {
      return this.attr('user.firstName') + ' ' + this.attr('user.lastName');
    }
  }
}
```

But this didn't: 

```
define: {
  firstName: 'Chris',
  lastName: 'Gomez',
  fullName: {
    value: function () {
      return this.attr('firstName') + ' ' + this.attr('lastName');
    }
  }
}
```

This PR fixes the typo that created the issue and expands on the existing dependent value generators test to include examples of reads without the dot-syntax. 

Closes https://github.com/bitovi/canjs/issues/1819

### DIFF
--- a/map/define/define.js
+++ b/map/define/define.js
@@ -61,16 +61,15 @@ steal('can/util', 'can/observe', function (can) {
 			originalGet = this._get;
 
 		// Overwrite this._get with a version that commits defaults to
-		// this.attr() as needed. Because calling this.attr() for each individual
-		// default would be expensive.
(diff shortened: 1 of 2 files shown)


daffl:
Awesome!

Issue #1762, opened 2015-06-22
Dynamic select menus don't live bind correctly


dylanrtt:
When creating a dynamic select menu with a stache ```#each``` such as in the following example, when the result of the rendered list changes, the selected value of the select menu changes to the first item, which isn't necessarily correct.

```html
<select can-value="{selectedOpt}">
  <option></option>
  {{#each opts}}
    <option name="{{name}}">{{name}}</option>
  {{/each}}
</select>
```

http://jsbin.com/fedixolaru/1/edit?html,js,output

This can occur when using a computed list or when using stache filtering.

justinbmeyer:
@dylanrtt Just found this myself.  Thanks for reporting.

akagomez:
Here's a demo of this being an issue with native DOM manipulation too: http://jsbin.com/tupodanihu/1/edit?html,js,output

justinbmeyer:
This is probably less likely to happen in minor because of the diffing.

dylanrtt:
#1827 fixed the case where the options change and the selected option remains in the list, but if the selected option is removed and added back to the list, the displayed option stays out of sync.

Here is the original JSBin updated to use `2.3.1`:
http://jsbin.com/hajupezufa/edit?html,js,output

daffl:
Should be closed now via 1401c401c49609f2d8d369d5950cdcaf139e6018

dylanrtt:
I don't think this is fixed. My previous comment is still true using master.

justinbmeyer:
@dylanrtt anyway you can simplify the test to something along the lines of: https://github.com/canjs/canjs/commit/1401c401c49609f2d8d369d5950cdcaf139e6018#diff-7bca9ce6f0378f28c17f2c02cd1eb9b0R1859 ?

justinbmeyer:
@dylanrtt how can I see the error?  Thanks.

justinbmeyer:
@dylanrtt is this how:

1.  Select "Alexis", "Curtis", "Curtis".
2.  Change the first select from "Alexis" to "Curtis" and then back to "Alexis"

?

dylanrtt:
http://jsbin.com/kagecetoho/edit?html,js,output

justinbmeyer:
@dylanrtt thanks!

justinbmeyer:
Checking out this problem a bit, it doesn't look like there is a "nice" solution.  Also, I'm not sure what the behavior should be.

## Two-way binding behavior

[...]

PR #1827, opened 2015-08-05
retain select value 

justinbmeyer:
For #1762, makes sure that when replacing option elements, the value is maintained.

### DIFF
--- a/view/bindings/bindings_test.js
+++ b/view/bindings/bindings_test.js
@@ -1275,5 +1275,53 @@ steal("can/view/bindings", "can/map", "can/test", "can/component", "can/view/mus
 		QUnit.equal(select.selectedIndex, 0, 'Got selected index');
 	});
 
-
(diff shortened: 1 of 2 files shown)


view/bindings/bindings_test.js
```diff
@@ -1275,5 +1275,53 @@ steal("can/view/bindings", "can/map", "can/test", "can/component", "can/view/mus
...
+		// the value is set asynchronously
+		setTimeout(function(){
+			ok(select.childNodes[1].selected, "value is initially selected");
+			
+			values(["7","2","5","4"]);
+			
+			ok(select.childNodes[1].selected, "after changing options, value should still be selected");
+			
+			
+			start();
```
akagomez: I always thought the `start()` should be called once the async bit had ended, and before the test. Is there a particular reason you did it this way, or does it not really matter?

akagomez:
+1

matthewp:
Looks like tests were failing because bower install failed; GitHub was probably down, so I restarted the tests.

PR #1836, opened 2015-08-06
Makes sure unbind is not call unnecessarily on can.Lists passed to can.view.live.list

### LINKED ISSUE #1835: can.view.live.list unbinds from the list unnecessarily
when can.view.live.list is called, it calls updateList which always calls teardownList

justinbmeyer:
fixes  #1835

### DIFF
--- a/view/live/live.js
+++ b/view/live/live.js
@@ -356,12 +356,9 @@ steal('can/util',
 					}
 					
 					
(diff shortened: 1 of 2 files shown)


akagomez:
I'm curious how this PR addresses the issue and it's not clicking for me. Would you mind helping me understand?

PR #1845, opened 2015-08-10
System Stache plugin: Add dynamic imports to bundle

matthewp:
This updates the Stache plugin to add dynamic `<can-import>` statements
to the System.bundle configuration, thus making this step unnecessary
for users if using can-import for dynamic loading.


gsmeets:
This seems to be doing some of the work that my can-stache-amd plugin does. Does this also work for amd based solutions?

matthewp:
No, Steal has a different plugin API than AMD.  In retrospect I wish we had made Steal fully AMD compliant and used the same plugin API.

PR #1849, opened 2015-08-12
Update system-config to reflect the current Steal env

matthewp:
Steal's System.env can now have multiple values such as:

```
System.env = "production,server";
```

This would apply on the server. This makes the `envs` config work. So
now there is a System.envMap which will be something like `{
server: true, production: true}`. So this change updates out
system-config to use that map.


matthewp:
Closing this in favor of using envs config.

PR #1868, opened 2015-08-27
Fixes #1862. Objects with addEventListener now bindable

### LINKED ISSUE #1862: can.bind should support objects with on and addEventListener methods
This is so that objects that support the EventTarget interface, and objects following the newer jquery convention of `on` can be used in templated event handlers in a can.Control, and more importantly, a can.Component.

We check for the `bind` function here:
https://github.com/bitovi/canjs/blob/master/util/jquery/jquery.js#L53

mjstahl:


daffl:
Do we know why the tests are failing? It looks like the AMD build isn't working.

mjstahl:
I have no idea what is happening.  Everything I change makes no difference to the tests passing.

PR #1877, opened 2015-09-02
Fixing FF debugging functions conflicting with Can Map and Event functions

tracer99:
FF introduced a debugging function called watch. This meant that if you had a property labeled watch, it would throw an exception.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 node_js: 0.10
 before_script: node_modules/.bin/bower install
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "canjs",
   "repo": "bitovi/canjs",
   "description": "Can do JavaScript MVC",
-  "version": "2.2.6",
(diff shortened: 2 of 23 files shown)


daffl:
This looks good. Two things: Could you squash your commits and can we add test that breaks in Firefox for this?

tracer99:
Arghh. It appears I made it worse. At this point my git skills have been exhausted.

daffl:
Oh no. I can look into it if you can provide a breaking test for Firefox. Thinking more about it I am not sure if always checking for `hasOwnProperty` is the right solution yet so I'd like to double check it.

tracer99:
https://gitter.im/bitovi/canjs?at=55e870c9d231aa8e5918fdbd Matthew agrees that this may not be the best solution either.

To have this fail in FF, simply change any of the tests you have already by replacing the key name with "watch" and it will fail in modern FF.

julia-allyce:
@tracer99 I am going to close this PR, and open an new issue for this. Would love to implement your contribution if you wanted to create a fresh PR with your fix and/or failing test.

PR #1878, opened 2015-09-02
1820 can map interrupt

joe-crick:
can.Map.interrupt plugin. Provides the ability to interrupt setting or removing the properties of a can.Map by wrapping them in a can.transaction. Will also supplant the batch functionality in setState for can.route to allow a route to be interrupted on set or remove.


daffl:
I'm closing this PR since we will extract it into a separate plugin as discussed.

PR #1879, opened 2015-09-03
Switch helper: get expression data in switch

### LINKED ISSUE #1857: switch/case might not work when nested
@daffl ran into a bug when putting together the guide.

matthewp:
This fixes #1857. The bug was that if the expression passed into switch
changed the switch helepr would not re-evaluate and instead on the
switches would re-evaluate. So you could never go from one case to the
default. This fixes it by setting up the binding to the expression
within the switch helper function, so that if the expression changes it
will reevaluate the entire switch.

### DIFF
--- a/view/stache/mustache_helpers.js
+++ b/view/stache/mustache_helpers.js
@@ -150,10 +150,11 @@ steal("can/util", "./utils.js","can/view/live",function(can, utils, live){
 			};
 		},
 		'switch': function(expression, options){
+			var expressionValue = resolve(expression);
 			var found = false;
 			var newOptions = options.helpers.add({
 				case: function(value, options){
-					if(resolve(expression) === resolve(value)) {
+					if(!found && expressionValue === resolve(value)) {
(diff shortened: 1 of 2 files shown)


view/stache/mustache_helpers.js
```diff
@@ -150,10 +150,11 @@ steal("can/util", "./utils.js","can/view/live",function(can, utils, live){
 			};
 		},
 		'switch': function(expression, options){
+			var expressionValue = resolve(expression);
 			var found = false;
 			var newOptions = options.helpers.add({
 				case: function(value, options){
-					if(resolve(expression) === resolve(value)) {
+					if(!found && expressionValue === resolve(value)) {
```
daffl: Did we decide to leave `!found` in?

matthewp: I don't know, I did because it seemed like if a case is already found at this point there is no reason to evaluate the value here. We can still talk about it though.

PR #1882, opened 2015-09-03
Add ability to override can.route's batch with other batching implementations

joe-crick:
Add ability to override can.route's batch with other batching implementations, like can.transaction


joe-crick:
Added test to show that the plugin works with can.route: https://github.com/canjs/can-interrupt/commit/dbc68580f29de6f8cb5064195b2f9360abaf8fc4

PR #1886, opened 2015-09-07
fixed regression component control destructor

gsmeets:
This fixes a regression introduced in #1803 where the component's
control destroy method isn't called anymore when the component is
removed from the DOM.


daffl:
Thank you. I don't know how I missed that. I will get a 2.2.8 out shortly.

gsmeets:
That would be great. I just updated my build scenario to grab npm packages during build, instead of checking them into source control, so I'm tied to release packages at the moment.

PR #1895, opened 2015-09-09
Improve sort plugin insert performance

akagomez:
**TLDR: Not ready for merge**

This PR is a work in progress that is being tracked in https://github.com/bitovi/canjs/issues/1883.


akagomez:
Benchmarks before: 

Adding items via .push([list]) x **0.02 ops/sec** ±0.95% (5 runs sampled) 
Adding items via .splice(0, 0, [list]) x **0.02 ops/sec** ±1.91% (5 runs sampled)
Adding items via .each([list]) + .push([item]) x **0.04 ops/sec** ±0.72% (5 runs sampled) 


Current: 

Adding items via .push([list]) x **4.89 ops/sec** ±2.12% (17 runs sampled) 
Adding items via .splice(0, 0, [list]) x **5.18 ops/sec** ±1.62% (17 runs sampled)
Adding items via .each([list]) + .push([item]) x **0.63 ops/sec** ±2.18% (6 runs sampled)

akagomez:
For speed, If multiple items are added in one call to `.push()` or `.splice()`, the native `.sort()` is called with the user's provided comparator. Then, each subsequent `_getInsertIndex` is masked to the remainder of the list. For instance:

```
list = ['a','c','e','g','i']
list.push('b','d','f');
    //-> _getInsertIndex('b'); // Search indexes 0,1,2,3,4; Returns: 1
    //-> _getInsertIndex('d', 1); // Search indexes 1,2,3,4,5; Returns: 3
    //-> _getInsertIndex('f', 3); // Search indexes 3,4,5,6; Returns: 5
```

Without this tactic the time to insert 20k items at once goes up by ~400ms.

The reason I mention this is that unfortunately I just learned that the native `.sort()` method's implementation can yield unreliable results. See [here](http://jsbin.com/xecogej/edit?js,console) and [here](http://blog.rodneyrehm.de/archives/14-Sorting-Were-Doing-It-Wrong.html#types).

This means that a bad native sort result can throw off the sort plugins much more reliable result because of the masking. It's a "will probably never happen" edge case, but still. 

The best compromise that I can think of is to keep track of the types as they're inserted. If the type changes while doing the inserts, re-sort the entire list (which will result in unnecessary swaps if something went wrong, but it will be corrected), then continue to insert items without the masking. 

Or maybe I just shouldn't rely on the native sort implementation.

akagomez:
Decided to remove the pre-sorting optimization. Was causing problems for even a simple case in PhantomJS. Lesson learned. DO NOT rely on native `.sort()` implementations. I should make note of this in the sort plugin docs.

daffl:
If and when this is wrapped up, could you re-submit it against minor since that is going to be the next release?

akagomez:
Re-opening against minor.

Issue #1202, opened 2014-08-13
mustache's "unless" doesn't seem to work with computes


ericcarino:
See my sample code below.  unless works just fine with booleans, but it doesn't appear to work with computes.  Using if/else works just fine.

````
<html>
    <head>
        <script id="template" type="text/mustache">
            {{#unless test1}}Inside unless block!{{/unless}}
            {{#if test1}}{{else}}Inside the else block!{{/if}}
            <p>I'm expecting the text above to be exactly the same with the text below.</p>
            {{#unless test2}}Inside unless block!{{/unless}}
            {{#if test2}}{{else}}Inside the else block!{{/if}}
        </script>

        <script src="http://ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.js"></script>
        <script src="http://canjs.com/release/latest/can.jquery.js"></script>
        <script type="text/javascript">
            $(function() {
                var data = new can.Map({
                    test1: false,
                    test2: can.compute(function() {
                        return false;
                    })
                });

                var template = can.view("#template", data)
                document.body.appendChild(template);
            });
        </script>
    </head>
    <body></body>
</html>
````

Output

````
Inside unless block! Inside the else block!

I'm expecting the text above to be exactly the same with the text below.

Inside the else block!
````

justinbmeyer:
@ericcarino 

Why are you setting a compute as a property value?

PR #1899, opened 2015-09-09
Fix issue with #unless helper and computes - #1202

DesignByOnyx:
A demo page showing the error can be found here: http://jsbin.com/pucucidofa/edit?html,js,output

The change essentially makes the `unless` helper a proxy to the `if` helper, but it flips the `fn` and `inverse` callbacks. All tests were updated to first expose the bug, and then the fix was implemented.


daffl:
If this should go into 2.3 can you resolve the conflicts and add it to the 2.3.0 milestone?

PR #1901, opened 2015-09-09
fixes #1891 by making executeAnonFuncs default to true

### LINKED ISSUE #1891: View bound to method breaks on re-evaluation
I couldn't easily find the wording to describe this issue - I think I lack the vocabulary. Fortunately the demo makes it pretty clear.

Actual:

```
Num: 2
Num (x2): function () { return this.attr('num') * 2; }
```

> Not working in latest: http://jsbin.com/zisuxiboga/edit?html,js,output

Expected:

```
Num: 2
Num (x2): 4
```

> Working in 2.2.1: http://jsbin.com/kisapixeve/edit?html,js,output

justinbmeyer:
can.compute.read will executeAnonymousFunctions to true, so even if it's not passed, that behavior is still present.

### DIFF
--- a/compute/read.js
+++ b/compute/read.js
@@ -1,14 +1,12 @@
 steal("can/util", function(can){
 	
 	
-	
-	
 	// there are things that you need to evaluate when you get them back as a property read
 	// for example a compute or a function you might need to call to get the next value to 
 	// actually check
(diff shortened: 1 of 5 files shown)


daffl:
This should probably go into master. It also is a regression in 2.2.6

PR #1904, opened 2015-09-10
Fixes #1790. Cast element.getAttrs to string in live.js

### LINKED ISSUE #1790: attr.split() causes errors in IE10,11 in live.js on CanJS 2.2.6 when using jQuery
Line 302 on live.js causes errors in IE10,11 (and presumably 9). The reason is that IE will format "attr" as an integer if the value is not a string:

 elements.getAttr(el, attributeName)   <-- cannot assume string in IE

The fix for it is simple:
Wrap the line in a String() cast.

Line 302 live.js: var attr = String(elements.getAttr(el, attributeName));

nlundquist:
Fixes #1790. Elements.getAttr correctly returns other types, but string is expected. Added test case that reproduces a non-string attribute being returned and subsequent error.

### DIFF
--- a/view/live/live.js
+++ b/view/live/live.js
@@ -638,7 +638,8 @@ steal('can/util',
 				can.data(wrapped, 'hooks', hooks = {});
 			}
 			// Get the attribute value.
-			var attr = elements.getAttr(el, attributeName),
+			// Cast to String. String expected for rendering. Attr may return other types for some attributes.
+			var attr = String(elements.getAttr(el, attributeName)),
 				// Split the attribute value by the template.
 				// Only split out the first __!!__ so if we have multiple hookups in the same attribute,
 				// they will be put in the right spot on first render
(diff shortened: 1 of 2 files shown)


view/live/live.js
```diff
@@ -638,7 +638,8 @@ steal('can/util',
 				can.data(wrapped, 'hooks', hooks = {});
 			}
 			// Get the attribute value.
-			var attr = elements.getAttr(el, attributeName),
+			// Cast to String. String expected for rendering. Attr may return other types for some attributes.
+			var attr = String(elements.getAttr(el, attributeName)),
```
daffl: Should we changes this to an implicit cast like `'' + elements.getAttr(el, attributeName)`

daffl:
The issue is slated for 2.2.8. Can you resubmit against master?

nlundquist:
Closing to resubmit in #1907.

Issue #1872, opened 2015-09-01
"for" attribute is "htmlFor" property


justinbmeyer:
This should be fixed here:

https://github.com/bitovi/canjs/blob/master/util/attr/attr.js#L41

justinbmeyer:
Do this in minor.

daffl:
Closed via #1920

Issue #1874, opened 2015-09-01
Conditional "readonly" attribute does not work


dylanrtt:
I can use a mustache condition block to make an element have a conditional `disabled` attribute but it does not work with `readonly`. It simply never applies the attribute regardless of the condition's truthiness.

http://jsbin.com/horehelute/edit?html,js,console,output

justinbmeyer:
Interesting.  I assumed it work would b/c `readonly` is configured here: https://github.com/bitovi/canjs/blob/minor/util/attr/attr.js#L47

justinbmeyer:
@alexisabril this is probably pretty simple.  Although the test should probably take place in can/util/attr.  I would start there to make sure it's doing what it should be doing.

DesignByOnyx:
`readonly` attribute maps to `readOnly` property - could be a start...

daffl:
Closed via #1920

PR #1920, opened 2015-09-15
#1872: map attribute 'for' to 'htmlFor' property

ilyavf:
In can.attr added a special attribute mapping 'for' to 'htmlFor' property. Tests added.


daffl:
Why is the attribute capital `READONLY`? I would probably lowercase it. Looks good besides that.

ilyavf:
It does not matter, we lower case it anyway.

ilyavf:
Closes #1872 and #1874.

Issue #1883, opened 2015-09-04
Sort plugin - Make adding items faster


akagomez:
- [x] ~~Use a binary search to find the `insertIndex` (currently we use selection sort)~~
- [x] ~~Implement the "add" api so that the binary search's `startIndex` can be configured (as an alternative to starting at the mid-point); Also, return the resulting `insertIndex`~~
- [ ] ~~Collect batched adds into a temporary list, sort the temporary list (this could use JS's native `.sort()`), then add them to the target list sequentially using the last adds `insertIndex` as the next adds `startIndex`~~

akagomez:
It's important to note that the sort plugin will still use selection sort for complete re-sorts since it results in fewer swaps.

akagomez:
Current `_getInsertIndex` (control): 

<img width="800" alt="screen shot 2015-09-08 at 3 30 26 pm" src="https://cloud.githubusercontent.com/assets/82166/9745959/d667290a-5642-11e5-9fa2-e975bfd8a73b.png">

New `_getInsertIndex` (binary search): 

<img width="800" alt="screen shot 2015-09-08 at 3 53 48 pm" src="https://cloud.githubusercontent.com/assets/82166/9745986/011c1796-5643-11e5-8f91-cefc0b868295.png">

akagomez:
Benchmarks before: 

Adding items via .push([list]) x **0.02 ops/sec** ±0.95% (5 runs sampled) 
Adding items via .splice(0, 0, [list]) x **0.02 ops/sec** ±1.91% (5 runs sampled)
Adding items via .each([list]) + .push([item]) x **0.04 ops/sec** ±0.72% (5 runs sampled) 


Current: 

Adding items via .push([list]) x **4.89 ops/sec** ±2.12% (17 runs sampled) 
Adding items via .splice(0, 0, [list]) x **5.18 ops/sec** ±1.62% (17 runs sampled)
Adding items via .each([list]) + .push([item]) x **0.63 ops/sec** ±2.18% (6 runs sampled)

akagomez:
The 3rd objective "Collect batched adds ..." can't actually be done. Even if batched, multiple calls to `.push()` won't get a `batchNum` (Duh!). The best way to benefit from batched inserts is to store all of your "push's" into an array and pass that array to push all at once, like: `list.push.apply(list, queuedInserts)`.

PR #1927, opened 2015-09-16
Optimize sort plugin inserts (#1883)

akagomez:
- Use binary search for _getInsertIndex
- Fix benchmarks.js
- Add benchmark(s)
- Add test

Notes associated with the original PR can be found here: #1895


test/benchmarks.js
```diff
@@ -26,7 +26,8 @@ steal("benchmark", function(){
 			return suite.on.apply(this, arguments)
 		}
 	};
-	steal().done().then(function(){
+
+	steal().then(function(){
```
daffl: I'm pretty sure this shouldn't use `.then`

PR #1936, opened 2015-09-17
List concat should not serialize items in the lists before adding them to the new concatenated list

### LINKED ISSUE #1616: can List concat serializes the objects in the list before adding it to the resulting list
Fiddle:
http://jsfiddle.net/qYdwR/2468/

when you click on a source item (either), it adds it to a bucket
I concat both buckets and check if the item is in either before adding it
the concat clones the items into a new array/list rather than just putting the item itself in the new array/list

the item should still be the same item after a concat
just also be in the new list/array

https://github.com/bitovi/canjs/blob/master/list/list.js#L998

Thoughts?

James0x57:
Tests added too

Fixes #1616

### DIFF
--- a/list/list.js
+++ b/list/list.js
@@ -740,9 +740,9 @@ steal("can/util", "can/map", "can/map/bubble.js","can/map/map_helpers.js",functi
 		concat: function () {
 			var args = [];
 			can.each(can.makeArray(arguments), function (arg, i) {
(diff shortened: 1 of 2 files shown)


James0x57:
Closing for now:
Travis CI is having issues and there's a point for discussion on this solution that I'll bring up in #1616

PR #1937, opened 2015-09-17
Adding dependency of can/route to can/view/href

### LINKED ISSUE #1861: can/view/href depends on can.route but does not import it
This is in minor

James0x57:
Closes #1861


daffl:
`can/view/href` has already been deprecated so we can leave it as an external plugin dependency.

PR #1940, opened 2015-09-18
Fixes #1939. Adding reduce method to can.List

### LINKED ISSUE #1939: can.List should implement a "reduce" method

mjstahl:
#1939

### DIFF
--- a/list/list.js
+++ b/list/list.js
@@ -770,6 +770,34 @@ steal("can/util", "can/map", "can/map/bubble.js","can/map/map_helpers.js",functi
 		},
 
 		/**
+		 * @function can.List.prototype.reduce reduce
+		 * @description The reduce() method applies a function against an accumulator and each
+		 * value of the array (from left-to-right) to reduce it to a single value.
+		 * @signature `list.reduce(callback[, initialValue])`
+		 * @param {function(previousValue, currentValue, index, list)} function to execute on each value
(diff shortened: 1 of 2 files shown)


list/list.js
```diff
@@ -770,6 +770,34 @@ steal("can/util", "can/map", "can/map/bubble.js","can/map/map_helpers.js",functi
...
+		 *      return p + c;
+		 * }); // -> 10
+		 *
+		 * list.reduce(function(p, c, i, l) {
+		 *      return p + c
+	     * }, 10); // -> 20
+		 * ```
+		 */
+		reduce: function(cb, initial) {
+			return Array.prototype.reduce.apply(this.serialize(), arguments);
```
justinbmeyer: This won't work in older browsers.

justinbmeyer:
Maybe move this to can/list/reduce and make it work for older browsers.

list/list.js
mjstahl: Yeah, I just went back and looked at the MDN docs.. and saw IE 9. _sigh_

julia-allyce:
@mjstahl whats happening with this PR?

justinbmeyer:
This shouldn't be added to 2.3.2.  It would be new functionality. It would need to be scheduled for a 2.4.

justinbmeyer:
This probably would work now in 3.0.  Please make this change in can-define.  Thanks!

Issue #1942, opened 2015-09-18
Can-event gets broken in production if method does not exist.


ilyavf:
```
<div can-click="nonExistingMethod"></div>
```

[This block] (https://github.com/bitovi/canjs/blob/master/view/bindings/bindings.js#L201) gets executed in development but is removed in production which causes an error in case a method does not exist.

In production because the line [206] (https://github.com/bitovi/canjs/blob/master/view/bindings/bindings.js#L206) is removed, execution goes further and line [259] (https://github.com/bitovi/canjs/blob/master/view/bindings/bindings.js#L259) `scopeData.value.apply` causes an exception because  `scopeData.value` is `undefined`.

daffl:
This has been fixed via #1943

PR #1943, opened 2015-09-18
Consistent behavior between dev and production

marshallswain:
Related to #1942.  This wraps only the dev warning in "steal-removes" so that production code behaves the same as development code when the first argument isn't available anywhere.

### DIFF
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -196,17 +196,17 @@ steal("can/util", "can/view/stache/mustache_core.js", "can/view/callbacks", "can
 
 				// We break out early if the first argument isn't available
 				// anywhere.
-
-				//!steal-remove-start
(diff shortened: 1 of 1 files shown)


daffl:
Shouldn't be the `return null;` be removed completely?

marshallswain:
That's basically what it was doing before.  Not having the return there was breaking our app, but only in production mode.

There's probably a better fix for the overall issue, but this at least provides consistent behavior in dev and production until it's worked out.

ilyavf:
@daffl If `return null` gets remove then [scopeData.value.apply](https://github.com/bitovi/canjs/blob/master/view/bindings/bindings.js#L259) will cause an exception since `scopeData.value` is undefined.

daffl:
No this is indeed the correct fix. Thanks!

PR #1945, opened 2015-09-19
#1944: on component destroy call its prototype's destroy to teardown

### LINKED ISSUE #1944: 2.3.0-pre.11: can.Component's inner Control not tearing down events
It looks like event teardown for can.Component events object/can.Control is broken in the latest can.  Upgrading from 2.2.6 to 2.3.0-pre.11 we are now seeing issues with our project.

These events, which are part of the page-level component, do not get torn down when the element is removed:
'{viewModel.appstate} globalSearchReady': function(){}
'{window} resize':function(){}

So every time you visit the page, you get duplicate handlers.

@ilyavf made a JsBin demonstrating the problem: http://jsbin.com/fobake/edit?html,js,console,output
If you resize the page (e.g. with showing/hiding HTML section) you will see the event being called twice.

ilyavf:
When component is removed its `.destroy` method didn't call `prototype.destroy` thus its `.off` method was not called. This fixes #1944.

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -507,6 +507,7 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/control", "ca
 			this._bindings.readyComputes = {};
 		},
 		destroy: function() {
+			can.Control.prototype.destroy.apply( this, arguments );
 			if(typeof this.options.destroy === 'function') {
 				this.options.destroy.apply(this, arguments);
 			}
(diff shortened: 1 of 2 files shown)


daffl:
This has already been fixed in 2.2.8 via #1886. The correct solution would've been to merge 2.2.8/2.2.9 master into minor.

PR #1954, opened 2015-09-25
Fixes #1938. Check for isArray, if not use toString method

### LINKED ISSUE #1938: can.isArray behaves differently using can.util.jquery vs can.util.domless
jQuery implements the following for `isArray` checks 

```javascript
Array.isArray
```
(delegates to native static method)

can.util.domless uses
```javascript
can.isArray = function (arr) {
        return arr instanceof Array;
    }
```

This causes odd behavior, in that checks against a class that extends `Array` return `false` for the first case and `true` for the second. IE

```javascript
 can.isArray = function (arr) {
        return arr instanceof Array;
 };
class MyArray extends Array {}
var a = new MyArray();
var b = [];

//using the above, assuming can.isArray was from can.util.domless

can.isArray(a) //true
can.isArray(b) //true

$.isArray(a) //false
$.isArray(b) //true
```
[codepen](http://codepen.io/anon/pen/YyqaGL)

I ran into this because I was making fancy extended arrays, adding tree utility methods etc..

Is this an issue for the libs that use can.util.domless? It seems that one could have a strange bug between a browser usecase and a domless usecase..

mjstahl:


daffl:
It looks like those tests are legitimately breaking. Probably in libraries other than jQuery.

PR #1959, opened 2015-09-30
Updating infrastructure to latest Node and Testee

daffl:
Uses engine-dependencies for Zombie. Tested on latest Node 4.1.1


daffl:
I will try this later.

PR #1965, opened 2015-10-01
Do not use pushstate routing on file:// protocol URLs

daffl:
Pushstate routing on `file://` URLs doesn't work so we shouldn't load it in that case. This is important for mobile and desktop apps that normally load from the file system. Originally tracked at https://github.com/donejs/donejs/issues/92


daffl:
I did verify manually but could not add a test since Browsers don't support XHR requests on the `file://` protocol.

Issue #1966, opened 2015-10-02
can-import: rename "as" to "export-as"


matthewp:
Going to keep it backwards compatible.

What this does in reality is export the property on the template's viewModel, so "export-as" is more appropriate than "as".

daffl:
Closed via #1967

PR #1967, opened 2015-10-02
Alias can-import `as` attribute to `export-as`

daffl:
`as=""` cause a lot of confusion. `export-as=""` is more clear. Coses #1966

### DIFF
--- a/view/stache/intermediate_and_imports.js
+++ b/view/stache/intermediate_and_imports.js
@@ -26,14 +26,14 @@ steal("can/view/stache/mustache_core.js", "can/view/parser",
 			attrStart: function( attrName ){
 				if(attrName === "from") {
 					inFrom = true;
-				} else if(attrName === "as") {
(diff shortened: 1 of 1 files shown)


view/stache/intermediate_and_imports.js
```diff
@@ -26,7 +26,7 @@ steal("can/view/stache/mustache_core.js", "can/view/parser",
 			attrStart: function( attrName ){
 				if(attrName === "from") {
 					inFrom = true;
-				} else if(attrName === "as") {
+				} else if(attrName === "as" || attrName === "export-as") {
```
matthewp: There's another line that needs this.

matthewp: This one: https://github.com/bitovi/canjs/blob/rename-as-1966/view/stache/intermediate_and_imports.js#L36

daffl: Duh, it's right there don't know why I missed that. Fixed now, thanks!

Issue #1971, opened 2015-10-06
expressions are caching Args computes in the wrong spots


justinbmeyer:
Something like:

```
{{#each things}}

  {{doSomething(.)}}
{{/each}}
```

breaks right now because the value for DOT (.) is being cached.  The caching needs to happen, but not the way it is happening.

daffl:
Looks like this has been fixed via #1972

Issue #1970, opened 2015-10-06
Event binding syntax: functions applied with the wrong context


matthewp:
Consider a template like:

```handlebars
<form ($submit)="doSearch()">...</form>
```

Where doSearch is a function on the component's viewModel.  The value of `this` in doSearch should be the component's viewModel instance. Instead it is `window`.  The problem is here: https://github.com/bitovi/canjs/blob/minor/view/bindings/bindings.js#L257

There is no scopeData.parent property, it's undefined.  I believe this should be `scopeData.scope._context`.

Will submit a breaking test in the morning.

justinbmeyer:
I've fixed this in my branch.  There's a few context bugs that I've found.  I'll push what I have tomorrow.

justinbmeyer:
I think this is fixed.

PR #1972, opened 2015-10-06
includes docs and fixes

justinbmeyer:
For #1970 and #1971 

This will require changes in can-connect/can/tag that are similar to the changes in can/view/href.


daffl:
I think the change you mean has been made in https://github.com/canjs/can-connect/pull/35 (at least the only difference I can see in can/view/href in this commit is https://github.com/bitovi/canjs/pull/1972/files#diff-c41f9367c9bc3372297f1cf3fc7f217fR20),

justinbmeyer:
It hasn't. Another change needs to be made. I changed how the .hash method works.

justinbmeyer:
https://github.com/canjs/can-connect/blob/263c195103926e89179764b9417d8e45ae447823/src/can/tag/tag.js#L98 needs to be changed.

PR #1974, opened 2015-10-06
Fixes #1969. Quotes 'case' property for IE8 compatibility

### LINKED ISSUE #1969: ie8 cannot use 'case' as a key
unless quoted: https://github.com/bitovi/canjs/blob/minor/view/stache/mustache_helpers.js#L156

mjstahl:


matthewp:
Thanks!

PR #1990, opened 2015-10-15
Fixes #1989. isArrayLike does not fail on non-object

### LINKED ISSUE #1989: isArrayLike needs to check for object type
When passing a non-object (e.g. boolean) `can.each` fails where it previously did nothing. This is due to the fix in c9a9a84a7e72ccea86939ae2f64f9e83feaf193b (using the `in` operator on a non-object will fail). The complete statement should be:

```js
var length = typeof obj === 'object' && 'length' in obj && obj.length;
```

mjstahl:
…where it previously did nothing.


daffl:
Gor Mark! CI failed because of JSHint though.

mjstahl:
@daffl Yeah I saw that this morning. I just haven't had time to fix it. I'll get it fixed some time today.

PR #2003, opened 2015-10-20
Fix wrong number of parameters when calling 'setup' in 'can/map/attributes'

meconcomputer:
Errors are encountered when using 'can/map/attributes' due to improper number of parameters used in can.Map/Model 'setup' wrapper


daffl:
Thanks! Just working on the 2.3 release so this should make it in pretty quick.

meconcomputer:
Thank you for a quick response :thumbsup:

Issue #2006, opened 2015-10-20
sort plugin: comparator set on init does not keep list sorted


alexisabril:
http://jsbin.com/wugefa/edit?html,js,console

Sort plugin works in 2.3.0-pre.1, but fails in 2.3.0-beta.6

alexisabril:
This is due to the removal of `delete this._init` on: https://github.com/bitovi/canjs/commit/8b333ab#diff-532d6324975f398d1901af12e9086953L110

Note: this._init has also been renamed to this._initializing.

Re-adding that line fixes this issue, however breaks https://github.com/bitovi/canjs/blob/minor/list/sort/sort_test.js#L451

PR #2011, opened 2015-10-21
Refactoring Construct.prototype._initializing to __.inSetup

alexisabril:
Allows for the triggering of events within a construct's init() method to preserve backwards compat.


alexisabril:
This is for issue #2006

daffl:
Was there a reason to rename the property instead of just moving it?

alexisabril:
Originally the property wrapped a construct's initialization status, however now it only wraps the setup. Variable name was changed to be more descriptive.

daffl:
Ok. Everything that uses that property should be in this repository so I think this is fine.

PR #2021, opened 2015-10-23
Adding scoped reference to 'can'. Removing dependancy on Global 'can'.

Psykoral:


list/promise/promise.js
```diff
@@ -1,4 +1,4 @@
-steal("can/list", function (list) {
```
Psykoral: In this case, 'list' was referenced but never used. 'can' was not referenced but used.

justinbmeyer:
Thanks!

Psykoral:
Fixing that last build error, give me a few

justinbmeyer:
I'm not sure what's breaking: https://travis-ci.org/bitovi/canjs/jobs/87143498#L2388

Anyway you an check it out?

The easy way is to run:

```
grunt testify
```

And then open `test/amd/jquery-2.html` and see if that's breaking in the browser.

Psykoral:
It looks like can.isPlainObject is defined in https://github.com/bitovi/canjs/blob/master/util/object/isplain/isplain.js#L29 and so the new local scoped can doesn't know what it is.

Psykoral:
:+1:  :100:

daffl:
Awesome, thank you!

Psykoral:
My pleasure!

PR #2036, opened 2015-10-28
Use regular 'routeUrl' helper syntax in can-href deprecation notice.

m-mujica:


matthewp:
I've been just using it like a normal helper:

```
href="{{routeUrl page='orders'}}"
```

justinbmeyer:
Yeah, the `normal helper` is what we ended up promoting now.

m-mujica:
@matthewp @justinbmeyer I like the 'regular helper' version better, I changed the deprecation notice to read like:

'Use the routeUrl helper like `href={{routeUrl prop=value}}`'

Is that better?

matthewp:
Yep, thank!

PR #2038, opened 2015-10-28
Allow relative can-imports

### LINKED ISSUE #2037: can/view/stache/system should normalize dynamicImport names
Otherwise steal-tools build/live-reload will fail when it tries to normalize them.

matthewp:
This fixes our can/view/stache/system plugin to normalize the dynamic
module names before adding them to System.bundle. This is needed because
steal-tools normalizes the bundle names and if the names are relative it
will throw an exception. Fixes #2037


matthewp:
Looking into why this is failing.

matthewp:
@daffl this is the PR i need in 2.3.1

PR #2041, opened 2015-10-29
Add can/map/define to default can require

### LINKED ISSUE #2040: Make define part of the core
Let's move can/map/define to be part of core canjs vs a separate plugin.

daffl:
This adds `can/map/define` by default when someone loads the `can` module with a module loader. Closes #2040


matthewp:
Shouldn't this be done in can/map itself?

daffl:
I was thinking about that, too but I think it should stay its own plugin (and you can't load it in can/map because they would depend on each other).

When we put everything into their own modules it might live with `can/map` and be included by default (but also give the option to require them separately).

justinbmeyer:
Yes, it should.

Issue #2044, opened 2015-10-30
Need equivalent of preventDataBindings to prevent component's bindings


matthewp:
Or maybe just reuse that when setting up component's bindings. I think this is needed to make `can-tag` work on can-imports.

matthewp:
Actually we can probably just pass that in as tagData.

justinbmeyer:
@matthewp is this closed?

matthewp:
It can be yes.

PR #2045, opened 2015-10-30
All can-import tests are passing again

### LINKED ISSUE #2042: Importing a template doesn't work with the new syntax
```
<can-import from='can/view/import/test/other.stache!' {^value}='*other' />{{> *other}}
```

Tries to import `*other.mustache`, not sure why yet.

matthewp:
This updates all of the can-import tests to use the new binding syntax.
One of the tests uses this method:

```handlebars
<can-import from='can/view/import/test/other.stache!' {^@value}='*other'
/>

{{{*other()}}}
```

While all of the others use the old syntax:

```handlebars
<can-import from='can/view/import/test/other.stache!' {^@value}='*other'
/>

{{> *other}}
```

So we know that both do work.

One change I had to make was to prevent can.Component from setting up
it's own bindings when can-import is used with `can-tag`. This is
because the bindings are performed in can/view/bindings instead.

Closes #2042 and #2044

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,3 +2,4 @@
 [*]
 end_of_line = LF
 indent_style = tab
+trim_trailing_whitespace = false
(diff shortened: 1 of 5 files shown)


matthewp:
@justinbmeyer I'd like you to review this PR, particularly the changes I made to `component/component.js` where I pass in a `preventDataBindings` flag into the tagData to fix the can-import can-tag problem. Maybe there's a more elegant way.

PR #2048, opened 2015-10-30
Added a setter for class attr to handle svg elements where className cann…

### LINKED ISSUE #2015: svgEl.className is readOnly so live binding does not work
SVG elements class name is readOnly so the only way to set it is to use
svgEl.setAttribute( "class", "..." );

This means that this:
&lt;svg class="{{#if cond}}liveBoundClassName{{/if}}">&lt;/svg>
does not work.

Alfredo-Delgado:
…ot be written to like `svg`. Resolves #2015

### DIFF
--- a/util/attr/attr.js
+++ b/util/attr/attr.js
@@ -38,7 +38,17 @@ steal("can/util/can.js", function (can) {
 			 * `element.getAttribute` and `element.setAttribute`.
 			 */
 			map: {
-				"class": "className",
+				"class": function(el, val) {
+					val = val || '';
+					el.className = val;
+
+					if(el.className !== val) {
+						// e.g. svg elements #2015
+						el.setAttribute('class', val);
+					}
+
+					return val;
+				},
 				"value": "value",
 				"innertext": "innerText",
 				"innerhtml": "innerHTML",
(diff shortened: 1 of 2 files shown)


justinbmeyer:
This shouldn't be reading className. I think it should just detect if the element is an svg element and set the attribute.

justinbmeyer:
Also, I think if someone set a className to `{toString: function(){ return "something"}}`, this would fail.

Alfredo-Delgado:
- I added tests for `{toString: function(){ return "something"}}` and they pass.
- The criterium for using `setAttribute` was updated to only apply to `SVGElement` instances.

util/attr/attr.js
```diff
@@ -38,7 +38,18 @@ steal("can/util/can.js", function (can) {
 			 * `element.getAttribute` and `element.setAttribute`.
 			 */
 			map: {
-				"class": "className",
+				"class": function(el, val) {
+					val = val || '';
+
+					if(el instanceof SVGElement) {
```
Alfredo-Delgado: Thinking of IE8, this won't work. I'll update to `if(el.namespaceURI === 'http://www.w3.org/2000/svg')` like [modernizr](https://gist.github.com/mstalfoort/1293822#file-gistfile1-html-L45).

Alfredo-Delgado: Also: http://jsperf.com/instanceof-vs-namespaceuri

PR #2059, opened 2015-11-05
Added an api page for `can.List.prototype.map`

Alfredo-Delgado:
Also fixed some copy pasta on `can.List.prototype.each`


moschel:
Test failure is unrelated, so merging

PR #2060, opened 2015-11-06
implementing compute to check for comparisons within #is. fixes #1931

### LINKED ISSUE #1931: eq helper called twice
Related to #1412.

If `eq` is used like:

```
{{eq foo.length bar.length}}
```

If foo.length and bar.length were to change together in a batch, and stay the same, this helper would be called twice and cause unnecessary updating of the DOM.

The solution is to create a `can.compute()` that represents if `.fn()` or `.inverse()` should be called.  For example:

```js
var lastValue, curValue,
	options = arguments[arguments.length - 1];

if (arguments.length - 2 <= 0) {
	return options.inverse();
}
var args = arguments;
var callFn = can.compute(function(){
  for (var i = 0; i < args.length - 1; i++) {
	curValue = resolve(args[i]);
	curValue = can.isFunction(curValue) ? curValue() : curValue;

	if (i > 0) {
		if (curValue !== lastValue) {
			return false;
		}
	}
	lastValue = curValue;
  }
  return true;
})


return callFn() ? options.fn() : options.inverse();
```

This way, `eq` is only called when the result of `callFn` changes.

justinbmeyer:
Fixes #1931

### DIFF
--- a/view/stache/mustache_helpers.js
+++ b/view/stache/mustache_helpers.js
@@ -115,25 +115,29 @@ steal("can/util", "./utils.js","can/view/live",function(can, utils, live){
 		},
 		'is': function() {
 			var lastValue, curValue,
-				options = arguments[arguments.length - 1];
+			options = arguments[arguments.length - 1];
 
 			if (arguments.length - 2 <= 0) {
 				return options.inverse();
 			}
 
-			for (var i = 0; i < arguments.length - 1; i++) {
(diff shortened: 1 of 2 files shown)


justinbmeyer:
@alexisabril can you merge this.  Also, please watch out for unnecessary spacing changes in your commits. Thanks!

julia-allyce:
@alexisabril  can you resolve your merge conflicts?

PR #2064, opened 2015-11-09
makes exporting function work, but not pretty

### LINKED ISSUE #2051: Cannot export a component's prototype function to the references scope
I should be able to export a function from a component like this:

```html
<foo-bar {^@foo}="*bar"></foo-bar>
{{*bar()}}
```

It *should* work for functions defined on the component's viewmodel or the viewmodel's prototype, but neither seems to work.

http://jsbin.com/zumebacuma/edit?html,js,console,output

justinbmeyer:
Closes #2051.  Fix is to use key parsing the right way.

### DIFF
--- a/view/bindings/bindings_test.js
+++ b/view/bindings/bindings_test.js
@@ -1464,6 +1464,9 @@ steal("can/view/bindings", "can/map", "can/test", "can/component", "can/view/mus
 	});
 	
 	test("exporting methods (#2051)", function(){
+		expect(2);
+		
+		
 		can.Component.extend({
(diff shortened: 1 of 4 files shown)


justinbmeyer:
@DesignByOnyx cc @matthewp looks like CI fail is due to can-simple-dom changes: https://travis-ci.org/canjs/canjs/jobs/90006250#L1998 .  I thought these had been resolved?

DesignByOnyx:
This is needed first: https://github.com/canjs/canjs/pull/2063

Issue #1800, opened 2015-07-22
Condition (helper) inside can-EVENT doesn't work


whitecolor:
in latest minor pre 2.3 code:

```HTML
<div can-click="{{#if canShowPopover}}togglePopover{{/if}}"></div>
```

causes an error:
```
Uncaught TypeError: Cannot read property 'compile' of undefined .../node_modules/can/view/stache/text_section.js 29 TypeError: Cannot read property 'compile' of undefined
```

it works in 2.2.6

justinbmeyer:
I don't think this is valid. We can fix it, but the #if should go on the outside of the whole attribute.

whitecolor:
Why not valid? It is usefull some times:

```HTML
<div can-click="{{#if shown}}hide{{else}}show{{if}}"></div>
```

outside `#if` doesn't work anyway yet.

whitecolor:
I think this may be related https://github.com/bitovi/canjs/issues/1770

whitecolor:
@daffl is it difficult to fix in minor?

daffl:
Doesn't a helper just get used like

```html
<div can-click="{{#canShowPopover}}togglePopover{{/canShowPopover}}"></div>
```

Also your example does not have a closing `{{/if}}`.

whitecolor:
@daffl it just a mistype, doesn't work any way

whitecolor:
Are you going to fix this soon?

matthewp:
Probably not soon, PRs welcome though.

whitecolor:
@matthewp not sure where to look it.

daffl:
I'll have a look next week. If this worked in 2.2 we should probably find out what changed.

justinbmeyer:
We turned off processing in can.view.attrs to make it so we can switch between single binding and two way binding.

beno:
This must have been something introduced between 2.2.6 and 2.2.7

My navigation component is now broken (I use SLIM and the ~ is short for {{ }})

    nav
     ul.nav.clearfix
       ~#actions
         li
           a href="#" can-click="{~.}" ~.

To avoid confusion, here's what this gets translated into:

    <nav>
      <ul class="nav clearfix">
        {{#actions}}
          <li><a can-click="{{.}}" href="#">{{.}}</a></li>
        {{/actions}}
      </ul>
    </nav>

This all works fine in 2.2.6, but fails in 2.2.7:

    scopeData.value.apply is not a function

[...]

PR #2066, opened 2015-11-09
Turns back on live-binding in custom attributes

### LINKED ISSUE #1770: Stache conditionals within custom attributes break in 2.3.0-pre.2
The following is possible in 2.2.6:

```html
<input type="text" can-value="{{#contact}}contact.{{/contact}}name">
```

This is helpful in situations where the context in which a template is rendered is dynamic. I am aware that can-value doesn't actually listen to changes to the attribute so it's not possible to change the field after the template is rendered, but it does work for making the initial render dynamic.

However, it seems to breaks the parser in 2.3.0-pre.2 (only version I tested since 2.2.6) on [view/stache/text_section.js#L29](https://github.com/bitovi/canjs/blob/v2.3.0-pre.2/view/stache/text_section.js#L29) and/or [view/stache/utils.js:32](https://github.com/bitovi/canjs/blob/v2.3.0-pre.2/view/stache/utils.js#L32). I don't know if there are plans to fix it before release but figured I would bring it up as soon as possible.

I have a feeling that some of you might say doing it this way is bad. Perhaps. If you guys do want to continue to support it, I can make a jsbin to reproduce the bug. If not, I can fix my code.

justinbmeyer:
For #1800.  Most likely fixes #1770.  Thanks @beno for the test.

### DIFF
--- a/view/bindings/bindings_test.js
+++ b/view/bindings/bindings_test.js
@@ -1491,16 +1491,21 @@ steal("can/view/bindings", "can/map", "can/test", "can/component", "can/view/mus
 	
 		var map = new can.Map({
 			actions: ["action1", "action2"],
-			action1: function(){console.log("item 1 clicked")},
-			action2: function(){console.log("item 2 clicked")}
+			action1: function(){
+				equal(calling, 0,"action1");
+			},
+			action2: function(){
+				equal(calling, 1,"action2");
+			}
 		});
 	
-		var frag = template(map);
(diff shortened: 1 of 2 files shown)


julia-allyce:
@justinbmeyer  it looks like there are failing tests with can-simple-dom :(

daffl:
Yes, I think I'll have to merge in latest master. Working on it.

PR #2086, opened 2015-11-15
Component doc

### LINKED ISSUE #1740: can.Component events prototype does not listen on all events in viewModel
Since can.Component.prototype.events is a can.Control 
https://github.com/bitovi/canjs/blob/master/component/component.js#L43
https://github.com/bitovi/canjs/blob/master/component/component.js#L377

and the lookup is set to the viewModel https://github.com/bitovi/canjs/blob/master/component/component.js#L465

i expect i can listen to all elements on the viewModel with the { } notation.

here is a JSBin that demonstrate the problem
http://jsbin.com/guxazuteba/7/edit

* for the Deferred there is no event fired at all.
* {foo} is changed by clicking, but no console log is seen
* {foobar}, which is "span", is only working

it would be also nice to listen on Deferred, for example:
```
events: {
    inserted: function(){

    },
    '{dates} resolved': function(){
      console.log('dates is resolved');
    }
  }
```

pYr0x:
* improve the component doc
* closes #1740
* add can.Component.prototype.simpleHelpers doc, hopefully corrent with the linking

improvement on my PR are welcome... :)

### DIFF
--- a/component/events.md
+++ b/component/events.md
@@ -50,6 +50,205 @@ Components have the ability to bind to special [can.events.inserted inserted] an
         }
       }
 
+## Listening to viewModel events
+
+As in [can.Control] we can listening for events inside or outside the Component. We can use a Templated Event Handler for this. Unlike [can.Control] the value inside `{Name}` are looked up on `viewModel` first followed by `this.options` and then the `window` object.
+You can use the events for:
+* Tracking changes on [can.List] or [can.Map] like `add`, `remove`, `change`, (`replaced`)
+* Tracking changes on properties on the viewModel
+* Tracking changes on [can.Model]-Instance like `created`, `updated`, `destroyed`
+* Tracking promises on [can.Deferred] 
+
+The following example shows all the described possibilities:
+
+    // set up some fixtures
+    can.fixture("GET /recipes",function(){
+        return [
+            {id: 1, name: "omelette"},
+            {id: 2, name: "pizza"}
+        ];
+    });
+
+    can.fixture("POST /recipes",function(){
+        return {id: 3, name: "hot dog"};
+    });
+
+    can.fixture("PUT /recipes/{id}",function(){
+        return {id: 3, name: "sushi"};
+    });
+    
+    
+    // describe a resource
+    var Recipe = can.Model.extend({
+        id: 'id',
+        findAll: 'GET /recipes',
+        create:  'POST /recipes',
+        update:  'PUT /recipes/{id}',
+    },{});
+
+
+    // the components viewModel
+    var ViewModel = can.Map.extend({
+        define: {
+            myMap: {
+                value: function() {
+                    return new can.Map({
+                        eggs: 2,
+                        tomatos: 3
+                    });
+                }
+            },
+            myList: {
+                value: function(){
+                    return new can.List({
+                        name: 'Marie'
+                    });
+                }
+            },
+            foo: {
+                value: "say hello"
+            },
+            bar: {
+                get: function(){
+                    return this.attr('foo').replace('foo', 'bar');
+                }
+            },
+            recipe: {
+                Type: Recipe
+            },
+            myAsyncData: {
+                // async getter
+                get: function(lastSet, deferred){
+                    this.attr("myPromise").always(deferred);
+                }
+            },
+            myPromise: {
+                Value: can.Deferred
+            }
+        }
+    });
+
+    can.Component.extend({
+        tag: "example",
+        template: '',
+        helpers: {},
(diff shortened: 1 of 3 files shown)


component/events.md
```diff
@@ -50,6 +50,205 @@ Components have the ability to bind to special [can.events.inserted inserted] an
...
+## Listening to viewModel events
+
+As in [can.Control] we can listening for events inside or outside the Component. We can use a Templated Event Handler for this. Unlike [can.Control] the value inside `{Name}` are looked up on `viewModel` first followed by `this.options` and then the `window` object.
+You can use the events for:
+* Tracking changes on [can.List] or [can.Map] like `add`, `remove`, `change`, (`replaced`)
+* Tracking changes on properties on the viewModel
+* Tracking changes on [can.Model]-Instance like `created`, `updated`, `destroyed`
+* Tracking promises on [can.Deferred] 
+
+The following example shows all the described possibilities:
```
justinbmeyer: This is a great example.  It might be a bit too long for someone to easily digest.  Do you think all these cases are necessary?  This seems to be documenting `can.Map` and `can.List`s events instead of how the "events" object works.

pYr0x: david meant https://github.com/canjs/canjs/issues/1740#issuecomment-149928218 this would be the best place.
since can.Controll isn't the future i think there would be a good place to show how you can listen to events thats fired in the viewModel.
this example indeet is long, but shows all possibilities. i think people can handle this

maybe we can add a reference to this documentation in can.List and can.Map (http://canjs.com/docs/can.Map.html#section_Listeningtochanges)

justinbmeyer: I should restate ... it's not so much about what people can handle (although I think that a 200 line example is a bit long to read)

... it's about having the right docs in the right places, especially API docs, and being able to maintain docs.  `Control` and `Component.prototype.events` are just "sugar" ontop of `can.Map.prototype.bind` and `can.List.prototype.bind`.  Those `.bind` functions should be the place where comprehensive docs about what events are possible are documented.  This makes sense because if we make changes to `.bind`, we will likely document it there.

The `Component.prototype.events` should just be explaining how the sugar maps to the underlying `.bind`.

component/events.md
```diff
@@ -50,6 +50,205 @@ Components have the ability to bind to special [can.events.inserted inserted] an
...
+            '{viewModel.recipe} created': function(response, event){
+                console.log("viewModel's 'recipe' created with ID: "+response.attr('id'));
+
+                // add a ID for this instance
+                this.viewModel.attr('recipe.id', response.attr('id'));
+                this.viewModel.attr('recipe.name','sushi');
+
+                // 7.2.)
+                // update
+                this.viewModel.attr('recipe').save();
```
justinbmeyer: The above code seems like it's doing something wrong.  What is this trying to show?  I can't think of a good case to assign an id of one Model instance to another model instance.

component/events.md
pYr0x: we have also http://canjs.com/docs/can.Control.html#section_Listeningtoevents for listening to events on a Control. 
this docs sould not explain how the Component.prototype.events works, it is an example how the event object can used for.

feel free to move the example to other docs...

component/events.md
pYr0x: you create an instance, get the ID back from the server. the response is just a response, not the recipe model. 
to update the recipe model an ID have to set. if not, save() would create a new recipe

component/events.md
```diff
@@ -50,6 +50,205 @@ Components have the ability to bind to special [can.events.inserted inserted] an
         }
       }
 
+## Listening to viewModel events
+
+As in [can.Control] we can listening for events inside or outside the Component. We can use a Templated Event Handler for this. Unlike [can.Control] the value inside `{Name}` are looked up on `viewModel` first followed by `this.options` and then the `window` object.
+You can use the events for:
+* Tracking changes on [can.List] or [can.Map] like `add`, `remove`, `change`, (`replaced`)
```
justinbmeyer: What is `replaced` for?  I think you mean when a reference to a list is changed?

pYr0x: yes, replaced is actual not a event.
so you are right, ```replaced``` means reference changed

pYr0x:
@justinbmeyer you can leave the compenent event., but we should add the simpleHelper documentation.

justinbmeyer:
@pYr0x anyway you can update this pull request to remove or shorten the big example and merge it into master?

daffl:
Since it is a documentation fix, I'll remove it from the milestone and we'll just merge it when it is ready.

justinbmeyer:
Closing for new 3.0 docs.

Issue #2077, opened 2015-11-12
Bug with conditional attribute


whitecolor:
In 2.3 (it worked in 2.2):
```HTML
{{#if preview}}next-page="{nextPage}"{{/if}}
```

If `preview` is `true` `nextPage` attribute value is string value `{nextPage}`, not actually value of `nextPage` in parent scope.

Maybe some why to do it with new binding sintax?

test is comming.

justinbmeyer:
You tried this with master?

whitecolor:
Yes with yesterday master.

whitecolor:
In this case:

```HTML
<some-comp {{#unless allowSome}}disabled{{/unless}}></some-comp>
```

if `disabled` attr is `htmlbool` in `some-comp` component, it event breaks with en erorr:

```
node_modules/can/view/stache/stache.js 301 Error: / is currently not supported within a tag.
```

justinbmeyer:
@whitecolor I get no errors with:

```
test("conditional attributes (#2077)", function(){
			can.Component.extend({
				tag: 'some-comp',
				viewModel: {
					define: {
						disabled: {
							type: 'htmlbool'
						}
					}
				}
			});
			var template = can.stache("<some-comp {{#unless allowSome}}disabled{{/unless}}></some-comp>");
			
			template();
		});
```

Can you submit a test?

justinbmeyer:
So the problem was that this was never working great to begin with.  However in 2.0, it was working on component initialization because attribute callbacks were called before tag callbacks.  

This meant that `next-page="{nextPage}"` was added to the element before the tag was created.  

Now this happens after.  

I'm trying to track down why we made this change, although I think it makes sense that components are invoked first.

justinbmeyer:
This commit changed each attribute to be added to the front: https://github.com/canjs/canjs/commit/bdb06d6a92e4b8de3a59d897fba03808f71cf80d

This has the effect of making the tag run last because it will be added first.

This was added because bindings needed the view-model to be created before they ran. This is no longer the case because can-component sets up bindings itself.

So the hack solution is to reverse this.  This will bring back compatibility.

But the better solution is to be able to handle attributes being added and removed.  

The better solution isn't easy to write because binding split between components and can/view/bindings.

I might make this work with can/view/bindings without components first, and then make components do it and try to share as much code as possible.

justinbmeyer:
[...]

PR #2089, opened 2015-11-17
refactor binding to updating bindings when attributes change, but also unifies binding behavior

justinbmeyer:
For #2077.

This refactor removes all binding logic from can/component and puts it in can/view/bindings.

This was done by making `makeDataBinding` able to get the `viewModel` mid-way through its processing.

The other big refactor was to create (and test) a `getBindingInfo` function.  If you give it an attribute node, it will tell you everything about the binding behavior:

```js
// Testing the old mustache behavior
var info = can.bindings.getBindingInfo({name: "foo", value: "bar"},{foo: "@"},"legacy");
deepEqual(info,{
	parent: "attribute",
	child: "viewModel",
	parentToChild: true,
	childToParent: true,
	childName: "foo",
	parentName: "foo",
	bindingAttributeName: "foo"
}, "legacy with @");
```

There's still a potential memory leak as non-viewModel bindings aren't tearing themselves down.  I'm probably going to make `makeDataBinding` always return a "post view model" continuation regardless if there's a viewModel needed or not.

However, I do want to make sure that we aren't creating a view model with `{($value)}="name"`.

### DIFF
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -553,7 +553,7 @@ steal("can/util", "can/view/stache/expression.js", "can/view/callbacks", "can/co
 								});
 							}
 						}
-					} 
+					}
 					// The parentCompute can sometimes be just an observable if the observable
 					// is on a plain JS object. This updates the observable to match whatever the
 					// new value is.
--- a/view/bindings/bindings.js
+++ b/view/bindings/bindings.js
@@ -4,7 +4,727 @@
 // for in template event bindings. These are usable in any mustache template, but mainly and documented
 // for use within can.Component.
 
-steal("can/util", "can/view/stache/expression.js", "can/view/callbacks", "can/control", "can/view/scope", "can/view/href", function (can, expression) {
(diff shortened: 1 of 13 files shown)


view/bindings/bindings.js
```diff
@@ -4,7 +4,727 @@
...
+			// the `viewModel` is created.
+			viewModel = makeViewModel(initialViewModelData);
+			
+			for(var i = 0, len = onViewModels.length; i < len; i++) {
+				onViewModels[i]();
+			}
+			
+			// Listen to attribute changes and re-initialize
+			// the bindings.
+			can.bind.call(el, "attributes", function (ev) {
```
pYr0x: so this should work ?
```
can.Component.extend({
  tag: "example",
  template: 'foobar is: {{foobar}}',
  viewModel: {},
  events: {}
});

var renderer = can.view('bar');

$('body').append(renderer({
  'name': 'hello'
}));
```
with 
```
<script id="bar" type="text/stache">
  <example {foobar}="{{name}}"></example>
</script>
´´´

justinbmeyer: Yes.

PR #2119, opened 2015-12-03
Fix can.view.txt hookup on self-closing tags

### LINKED ISSUE #2113: can.view.txt does not handle binding hookup on self-closing elements correctly
This issue surfaces when a hookup ends up inside a `colgroup` element. Via the `elements.tagMap` a `colgroup` element supports only a `col` child-element, however `col` is self-closing and may not hold content.

If you have a live-binding expression inside a `colgroup` (e.g. to render a variable amount of columns in a table), then the hookup in [view/render.js;207-222](https://github.com/canjs/canjs/blob/master/view/render.js#L207-L222) will emit HTML for a tag with content, e.g. `<col data-view-id="1234">@@!!@@</col>`.

HTML5 compliant browsers will coerce this to a self-closing col element and will push the `@@!!@@` marker outward: placing it before the containing `table` element, where it will no longer be replaced when hookups are resolved and will end up being exposed to the user.

The solution is to retain some special knowledge of self-closing tags and update the hookup to emit the correct self-closing HTML that way.

rjgotten:
Fixes #2113 

In some cases hookup is done by can.view.txt on tags that are self-closing.

The default hookup logic always renders tags with content, which is bad because browsers attempt to fix this when parsed into a DOM fragment and end up bubbling the content outside of the node, leaving ugly @@!!@@ marker text everywhere.

This fixes that by adding a few checks for self-closing tags and rendering appropriate hookup placeholder markup.


matthewp:
Thanks! Can you add a test?

rjgotten:
> Thanks! Can you add a test?

Done.

rjgotten:
@matthewp 
Looks like PhantomJS and *only* PhantomJS is failing on part of the test I added.
When I ran it in an actual browser, the test passed correctly.

view/view_test.js
```diff
@@ -873,4 +873,30 @@ steal("can/view/callbacks",
...
+			"mustache",
+			"stache"
+		], function (ext) {
+			var
+				frag = can[ext](templates[ext])({ list : list }),
+				div  = document.createElement("div");
+
+			div.appendChild(frag);
+
+			equal(div.querySelectorAll("col").length, 2, "Hookup with self-closing tag rendered properly with " + ext );
```
matthewp: Use `getElementsByTagName` instead here.

rjgotten:
@matthewp 
> Use `getElementsByTagName` or `can.$` instead here.

Still fails, I'm afraid.

justinbmeyer:
> When I ran it in an actual browser, the test passed correctly.

How are you running it in an actual browser?

matthewp:
@rjgotten Looks like it's failing in the dist tests. These are the tests that use the standalone build. You should be able to replicate in a browser by running `grunt build` and then opening up test/dist/mootools.html

daffl:
It looks like it is failing for non-jQuery libraries. You can run those by going to `canjs/test/index.html` or `canjs/test/<name>.html`.

rjgotten:
@justinbmeyer 
> How are you running it in an actual browser?
Opening the html file by hand from a test server instance, pretty much following what @matthewp is writing as well.

@daffl, thanks for pointing me at the non-jQuery libraries. I'll have a look tomorrow when I have access to my dev workstation that has everything set-up.

rjgotten:
@matthewp
Found it.

The problem is not in the patch code or in the unit test.

The unit test is exposing an underlying weakness in the fallback `can.buildFragment` implementation. That implementation currently fails to account for some of the table-related special cases that jQuery's version of the logic does account for, among which are free `colgroup` and `col` elements.

matthewp:
Thanks for doing the digging!  We can get this in 2.3.6.

daffl:
Awesome. Looks good to me. Also verified that `querySelectorAll` is fine for IE8.

PR #2143, opened 2015-12-15
Do not fire change event on NaN to NaN compute updates.

### LINKED ISSUE #2128: can.compute | NaN -> NaN dispatches a change event
https://github.com/canjs/canjs/blob/master/compute/proto_compute.js#L384

Currently `can.compute` fails to take into account the fact that `NaN !== NaN`, which causes computes to dispatch change events when their computed value remains `NaN`,

m-mujica:
Closes #2128.

screen shot 2015-12-15 at 18 22 22

screen shot 2015-12-15 at 18 22 41

### DIFF
--- a/compute/compute_test.js
+++ b/compute/compute_test.js
@@ -958,4 +958,14 @@ steal("can/compute", "can/test", "can/map", "steal-qunit", function () {
 		
 	});
 
+	test("compute should not fire event when NaN is set multiple times #2128", function() {
+		var compute = can.compute(NaN);
+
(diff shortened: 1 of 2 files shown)


daffl:
The tests are legitimately breaking. For some reason this change causes a lot of other issues.

justinbmeyer:
@m-mujica seems like there's a problem with this.  Ping me tomorrow and lets take a look.  (I'll be around tonight if you want to look at it too). Thanks!

PR #2148, opened 2015-12-20
Firefox debugging functions conflicting with can.Map

### LINKED ISSUE #2069: FF debugging functions conflicting with can.Map
This is from PR #1877 
"FF introduced a debugging function called watch. This meant that if you had a property labeled watch, it would throw an exception."
cc: @tracer99

prashantsharmain:
Gecko exposes methods named "watch" and "unwatch" on Object.prototype. If a can.Map doesn't define these attributes, the ___get method in map.js incorrectly treats these as computed attributes. Fix it by ignoring these properties before invoking the compute function on the attribute.

Fixes #2069


prashantsharmain:
The build failure looks unrelated to the fix I made. I'm going to close the PR and then open it again. I expect this would trigger the Travis CI build again.

PR #2153, opened 2015-12-29
fixing comments in 'Extending a Map' example

apieceofbart:
I noticed small error while going through tutorial. I'm new to github so I'm not sure if this is correct way to submit a pul request.


daffl:
Awesome, thank you!

PR #2157, opened 2016-01-01
VDOM - append text node for script tag innerHTML

### LINKED ISSUE #2156: vdom innerHTML implementation should append a text node for script tags
The innards of the `innerHTML` implementation in vdom does this:

    ...
    var frag = parser.parse(""+html);
    this.appendChild(frag);

However, script tags contents are CDATA and should not be parsed as html.  Instead, a text node should just be appended with whatever content is given to it.

DesignByOnyx:
Closes #2156

### DIFF
--- a/util/vdom/vdom.js
+++ b/util/vdom/vdom.js
@@ -19,7 +19,12 @@ steal("can/util/can.js", "can-simple-dom", "./build_fragment/make_parser", funct
 						cur = this.firstChild;
 					}
 					if(""+html) {
-						var frag = parser.parse(""+html);
+						var frag ;
+						if (this.nodeName === "SCRIPT") {
+							frag = document.createTextNode(""+html);
+						} else {
+							frag = parser.parse(""+html);
+						}
 						this.appendChild(frag);
 					}
 				}


DesignByOnyx:
@daffl - this fixes an issue on our current project and it would be greatly appreciated to get this in. Please let me know if there are any hangups.

daffl:
Can this be tested?

DesignByOnyx:
There are no tests running for vdom at the moment - so I didn't write one. I can verify that this works on a client project. If you were to set up the testing structure for vdom, I would write a test to cover this. I just don't have the time or authority to set it up and determine whether these tests should run in a worker or not and all the discussion therein.

matthewp:
@DesignByOnyx All of the stache tests run both in the regular (real) DOM and using can/util/vdom.

DesignByOnyx:
Interesting. then can you suggest where I might be able to test this? The test would look something like this:

```js
test("Script contents are not parsed in vdom", function () {
    var script = document.createElement("script");
    var code = "var markup = '<Foo';";
    try {
        script.innerHTML = code;
        assert.equal(script.innerHTML, code, "Script content set correctly.");
    } catch () {
        assert.ok(false, "Failed to set script contents");
    }
});
```

PR #2168, opened 2016-01-06
Benchmark array diff utility

akagomez:
I created some benchmarks of the current array diff utility in CanJS. I was pretty surprised to see that even with 10k items in the two compared lists, the runtime is still less than 1ms. Am I reading these results correctly? 

```
SUITE diff
  10k - same x 30,860 ops/sec ±1.26% (61 runs sampled) 
    [Average runtime: 0.032403886673781045]
  10k - slightly different x 21,544 ops/sec ±1.19% (58 runs sampled) 
    [Average runtime: 0.04641671862046386]
  10k - very different x 1,379 ops/sec ±1.77% (56 runs sampled) 
    [Average runtime: 0.7251001082251083]
```

Now, I will say it doesn't take much for it to defer to a complete replacement, but if all you've got is simple adds/removes this is pretty impressive.

### DIFF
--- a/util/array/benchmark.html
+++ b/util/array/benchmark.html
@@ -1,0 +1,1 @@
+<script src="../../node_modules/steal/steal.js" main="can/util/array/diff_benchmark"></script>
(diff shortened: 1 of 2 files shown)


akagomez:
@justinbmeyer Can you confirm I got this right? I double and triple checked but it's one of those "too good to be true" situations.

PR #2170, opened 2016-01-07
Allow anchor tags to be children of inline elements and contain block elements

### LINKED ISSUE #2169: Parser is not allowing anchor tags as child of inline elements
I have content which looks like this:

    <p><span>Some text with an <a>anchor</a> in it</span></p>

but since the anchor is defined as a block element, it is getting converted to this:

    <p><span>Some text with an </span><a>anchor</a> in it</p>

The anchor element is an inline element, but per html5 spec it's the only inline element which allows block level children. The parser should take this into consideration. The "a" tag should be part of the "inline" set of elements ([not the "block" group](https://github.com/canjs/canjs/blob/master/view/parser/parser.js#L53)), and conditional logic should be used to prevent the "a" tag from being closed when a block element is encountered.

DesignByOnyx:
Fixes #2169 

This also preserves the ability to have block elements as children of an anchor - there is still a test for that still passes.

### DIFF
--- a/view/parser/parser.js
+++ b/view/parser/parser.js
@@ -51,11 +51,11 @@ steal(function(){
 	var empty = makeMap("area,base,basefont,br,col,frame,hr,img,input,isindex,link,meta,param,embed");
 
 	// Block Elements - HTML 5
-	// a is traditionally inline, but should allow block-level elments inside it, so it should be treated like a block-level element when parsed
+	// For an INLINE element which can have BLOCK children, include that element in BOTH lists
 	var block = makeMap("a,address,article,applet,aside,audio,blockquote,button,canvas,center,dd,del,dir,div,dl,dt,fieldset,figcaption,figure,footer,form,frameset,h1,h2,h3,h4,h5,h6,header,hgroup,hr,iframe,ins,isindex,li,map,menu,noframes,noscript,object,ol,output,p,pre,section,script,table,tbody,td,tfoot,th,thead,tr,ul,video");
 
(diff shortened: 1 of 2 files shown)


justinbmeyer:
I don't think a should be special cased, we probably need a variable that contains other elements that could be treated like an 'a'.

DesignByOnyx:
That's a longer conversation. All I am trying to fix the the fact that special treatment was already being given to the anchor element - and that treatment was broken. If we really want to replicate html then we stop auto-closing tags altogether - but that's likely to break implementations.

I am needing this on a current project and have tested this change on that project (in addition to the unit tests written in this PR)

justinbmeyer:
@DesignByOnyx I don't I was being clear.  

```
while (last && inline[last] && last !== "a") {
```

Having "a" special cased like this vs something more like:

```
var blockAndInline = {"a", true}
``

changes the while loop to:

```
while (last && inline[last] && !blockAndInline[last]) {
```

justinbmeyer:
>  If we really want to replicate html then we stop auto-closing tags altogether 

That actually wouldn't be replicating HTML.   Replicating HTML would involve more complex parsing that was better able to tell when to auto-close tags.  HTML does auto-close tags.

The CI is erroring.  If you'd like to see this in 2.2.9, please remove the special casing from the while loop and get Travis to pass.

justinbmeyer:
Even better ... I wonder if there's a way to include 'a' in both the block and inline list and have the logic work.

PR #2177, opened 2016-01-09
makes jQuery work without public buildFragment and overwritable domManip

### LINKED ISSUE #2176: jQuery 2.2 compatible.
jQuery 2.2 was just released without a public .buildFragment.  We need to support it.

justinbmeyer:
Closes  #2176 

This makes jQuery 1.12 and 2.2 work.


mjstahl:
+1 locally

mjstahl:
fixed merge conflicts.

rjgotten:
@justinbmeyer 

Why not keep your old `buildFragment` logic and feature-detect the existence of the `jQuery.parseHTML( data [, context ] [, keepScripts ] )` method, which is part of the stable public API?

The `parseHTML` method does everything `$.buildFragment` did, but it flushes the contents of the document fragment that it internally uses with `$.buildFragment` into a plain array of DOM nodes and returns that to the caller. (Note: DOM *nodes*; not elements, so it still contains text nodes, comment nodes, etc. as well.)

Since you were cloning the cached fragment into a fresh fragment anyway, all you need to do with that array of nodes is pour it into a new empty document fragment of your own creation, and you're done.

justinbmeyer:
@rjgotten I thought about that, but knew I had the `buildFragment` that worked everywhere else.  Also, I wanted to avoid another "convert to a fragment" step.  Meaning, that yes, we were creating a fresh fragment (via DOM clone method), but my thought that `.clone()` would be much faster than having `parseHTML` break the DOM into an array, and then us re-assemble it into another fragment.  `clone()` presumably doesn't have to keep updating `.nextSibling` references.

We can maybe check if the performance is about the same.

justinbmeyer:
@mjstahl what did you change when you fixed the merge conflicts?  It seems things are breaking now.  Is there a merge commit?  


Also, I noticed that the `I'm also seeing a `view/stache/test/system/test.html` file looks different than the standard file.  Where did you get that structure from?  steal-qunit should be imported, not loaded via script tags.

justinbmeyer:
@mjstahl this is what I see when I open that page.

localhost_8080_canjs_view_stache_test_system_test_html

justinbmeyer:
Ah, it seems that wasn't intended to be opened as a standalone test page. I'm going to rename it.

mjstahl:
@justinbmeyer AllI did was resolve the merge conflicts with `package.json` and `bower.json`

PR #2178, opened 2016-01-10
Update package.json

pYr0x:
should'n we update steal and steal-tools to 0.13?


mjstahl:
@pYr0x If all goes well with #2175 tests this will be updated.

pYr0x:
close this @mjstahl has updated to steal 0.13
good job

PR #2221, opened 2016-01-29
Fixing compute event dispatching to properly evaluate nested templates

### LINKED ISSUE #1866: Child bindings are called before the parent in Stache
In the following example, if I start on `foo/view` and route to `bar/edit`, the template code corresponding to `foo/edit` is inserted temporarily... somehow after `bar/edit` is, even though I see `editing: bar` on the screen at the end...?

```html
{{#eq page 'foo'}}
  {{#eq action 'view'}}
    viewing: foo
  {{/eq}}
  {{#eq action 'edit'}}
    editing: foo
  {{/eq}}
{{/eq}}

{{#eq page 'bar'}}
  {{#eq action 'view'}}
    viewing: bar
  {{/eq}}
  {{#eq action 'edit'}}
    editing: bar
  {{/eq}}
{{/eq}}
```

This can be a serious problem if a complex component is being inserted that does things you potentially do not want to happen. As an example, I was getting the bug from #1865 because a component in the `foo/edit` part of the page was exporting its view model, which was really confusing...

http://jsbin.com/woqumidivi/2/edit?html,js,console,output

daffl:
Closes #1866

### DIFF
--- a/compute/get_value_and_bind.js
+++ b/compute/get_value_and_bind.js
@@ -12,7 +12,7 @@
 // - can.__observe - All other observes call this method to be visible to computed functions.
 // - can.__notObserve - Returns a function that can not be observed.
 steal("can/util", function(can){
-	
+
 	function ObservedInfo(func, context, compute){
 		this.newObserved = {};
 		this.oldObserved = null;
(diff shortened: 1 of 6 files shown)


daffl:
Argh, I don't know where all the whitespace diffs are coming from, `.editorconfig` is set up not to do that.

PR #2222, opened 2016-01-29
Add the ability to inherit values through the define plugin

### LINKED ISSUE #1322: Ability to use map.define on all Map classes in inheritance chain
At the moment it's not possible to use map.define plugin to define properties both on parent and child Map instances. Define config in child class completely overrides parent's "define" section. It should extend instead.

Please check the example: http://jsfiddle.net/njuw9sss/1/

chasenlehara:
This adds the ability to have two maps, one inheriting from the other, that both use the define plugin.

Fixes #1322

### DIFF
--- a/map/define/define_test.js
+++ b/map/define/define_test.js
@@ -1136,4 +1136,125 @@ steal("can/map/define", "can/route", "can/test", "steal-qunit", function () {
 			start();
 		}, 200);
 	});
+	
+	test("can inherit computes from another map (#1322)", 4, function(){
+		var string1 = 'a string';
+		var string2 = 'another string';
+
+		var ParentMap = can.Map.extend({
+			define: {
+				inheritedProp: {
+					get: function() {
+						return string1;
+					},
+					set: function(newVal) {
+						equal(newVal, string1, 'set was called');
+					}
+				},
+				parentProp: {
+					get: function() {
+						return string1;
+					}
+				}
+			}
+		});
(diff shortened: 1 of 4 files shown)


marshallswain:
Sounds cool.  Did you make docs so I can see how to use it?

chasenlehara:
@marshallswain, yes, just added docs: https://github.com/canjs/canjs/pull/2222/files#diff-ea73bd9b162a471eb881dae973216353R113

mickmcgrath13:
Can you use super?
```javascript
   var ChildMap = ParentMap.extend({
      define: {
        inheritedProp: {
          get: function() {
            return this._super.apply(this,arguments) + 1;
          }
        }
      }
    });
```

chasenlehara:
Hey @mickmcgrath13, it doesn’t work out of the box with the super plugin, but I started working on making `_super` available and it looks like I’ll be able to do it. I’ll continue working on it and update my PR.

justinbmeyer:
@chasenlehara this would be very nice for 3.0.  Any chance you get get a day or so to make it happen?

chasenlehara:
@justinbmeyer I have some other open source stuff I want to work on, so someone can take these issues if I don’t get to them first: https://github.com/canjs/can-map-define/issues/2 & https://github.com/canjs/can-define/issues/18

chasenlehara:
Closing this because it won’t be merged into a release.

Issue #2164, opened 2016-01-04
deprecate can/control/route ~2


justinbmeyer:
- [x] - remove it from major
- [ ] - give a warning in master

PR #2225, opened 2016-02-01
deprecates can/control/route warning 

justinbmeyer:
#2164


pYr0x:
@justinbmeyer can we merge that? 
i think we should merge all acceptable doc PR so that we can start building can 3.0 docs

PR #2232, opened 2016-02-02
2190 canjs api update

mjstahl:


view/doc/elements_and_attributes.md
```diff
@@ -36,7 +36,7 @@ With the release of canJS 2.3 the new [can.view.bindings binding syntax] provide
       
     * One-Way binds the `value` in the [can.view.Scope scope] to a element's attribute or property.
       ```
-      <input {$placeholder}="value">
+      <input {placeholder}="value">
```
pYr0x: why remove the $ sign?

http://canjs.com/docs/can.view.bindings.toChild.html

```
{$child-prop}="key"

Imports key in the scope to childProp property or attribute on the element.
```

PR #2247, opened 2016-02-09
Deprecate can.isDeferred. More accurate can.isPromise.

### LINKED ISSUE #1747: Better promise detection
A browser's Promise will fail can.isPromise.  I think we should change the test to something like:

```
can.isPromise = function(obj){
  return obj && (
     ( window.Promise && (obj instanceof Promise) ) ||
    ( can.isFunction(obj.then) && can.isFunction( obj["catch"] || obj.fail )
  )
}
```

### LINKED ISSUE #2359: Stache promise support doesn't work with ES6 promises
https://canjs.com/docs/can.stache.html#section_WorkingwithPromises

The support described here for promises in stache templates appears to only apply to jQuery deferreds. ES6 promises are just treated as any other object in a template.

If this is intended the docs should be updated to reflect it. 

A test case to demonstrate: https://github.com/nlundquist/canjs-template-promise-test

Correction: this is resolved by #2247

nlundquist:
Resolves #1747, closes #2359

isPromise needs to explicitly reject can.List as it shouldn't be handled as a promise and the can.List promise plugin is likely to be removed in a coming release.

### DIFF
--- a/test/pluginified/2.0.5.test.js
+++ b/test/pluginified/2.0.5.test.js
@@ -316,7 +316,7 @@ var __m1 = (function () {
 				update: function () {
 					var deferred = this.scope.attr('deferreddata'),
 						scope = this.scope;
-					if (can.isDeferred(deferred)) {
+					if (can.isPromise(deferred)) {
 						this.scope.attr('waiting', true);
 						deferred.then(function (items) {
 							scope.attr('items')
@@ -2105,7 +2105,7 @@ var __m32 = (function () {
 			ob = new can.Map({
 				test: dfd
 			});
-		ok(can.isDeferred(ob.attr('test')), 'Attribute is a deferred');
+		ok(can.isPromise(ob.attr('test')), 'Attribute is a deferred');
 		ok(!ob.attr('test')
 			._cid, 'Does not have a _cid');
 	});
@@ -4044,13 +4044,13 @@ var __m39 = (function () {
 			bar: bar
 		};
 		stop();
-		ok(can.isDeferred(original.foo), 'Original foo property is a Deferred');
+		ok(can.isPromise(original.foo), 'Original foo property is a Deferred');
 		can.view(can.test.path('view/test//deferred.ejs'), original)
 			.then(function (result, data) {
 				ok(data, 'Data exists');
 				equal(data.foo, 'FOO', 'Foo is resolved');
 				equal(data.bar, 'BAR', 'Bar is resolved');
-				ok(can.isDeferred(original.foo), 'Original property did not get modified');
+				ok(can.isPromise(original.foo), 'Original property did not get modified');
 				start();
 			});
 		setTimeout(function () {
--- a/test/pluginified/2.0.5.test.js
+++ b/test/pluginified/2.0.5.test.js
@@ -316,7 +316,7 @@ var __m1 = (function () {
 				update: function () {
 					var deferred = this.scope.attr('deferreddata'),
 						scope = this.scope;
-					if (can.isPromise(deferred)) {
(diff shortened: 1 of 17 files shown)


util/can.js
```diff
@@ -15,9 +15,16 @@ steal(function () {
...
-		return obj && typeof obj.then === "function" && typeof obj.pipe === "function";
+	
+	can.isDeferred = function() {
+		can.dev.warn('can.isDeferred: this function is deprecated and will be removed in a future release. can.isPromise replaces the functionality of can.isDeferred.');
+		return can.isPromise.call(this, arguments);
+	};
+	can.isPromise = function(obj){
+		return obj && (
+			(window.Promise && (obj instanceof Promise)) ||
+			(can.isFunction(obj.then) && can.isFunction(obj.catch || obj.fail) && !(obj instanceof can.List))
```
matthewp: I don't think checking for catch/fail is right as this is not a required to be a Promise.  This is probably the way to do it: https://github.com/then/is-promise/blob/master/index.js

nlundquist: Fair enough. Looking at the diffferences to what you linked, you think I should just remove the isFunction(catch || fail)? Or should I also check if the argument is an object or function?

matthewp: Not sure, wondering what scenario it is a function.  Should this work with jQuery's deferreds?  If so I'd make sure that it does (think it should)

util/can.js
```diff
@@ -15,9 +15,16 @@ steal(function () {
...
-	can.isDeferred = can.isPromise = function (obj) {
-		// Returns `true` if something looks like a deferred.
-		return obj && typeof obj.then === "function" && typeof obj.pipe === "function";
+	
+	can.isDeferred = function() {
+		can.dev.warn('can.isDeferred: this function is deprecated and will be removed in a future release. can.isPromise replaces the functionality of can.isDeferred.');
+		return can.isPromise.call(this, arguments);
+	};
+	can.isPromise = function(obj){
+		return obj && (
```
matthewp: Make this be `!!obj` so it returns `false` when provided `undefined`. This function should always return a boolean.

util/can.js
```diff
@@ -15,9 +15,16 @@ steal(function () {
 	// An empty function useful for where you need a dummy callback.
 	can.k = function(){};
 
-	can.isDeferred = can.isPromise = function (obj) {
-		// Returns `true` if something looks like a deferred.
-		return obj && typeof obj.then === "function" && typeof obj.pipe === "function";
+	
+	can.isDeferred = function() {
+		can.dev.warn('can.isDeferred: this function is deprecated and will be removed in a future release. can.isPromise replaces the functionality of can.isDeferred.');
+		return can.isPromise.call(this, arguments);
```
matthewp: This changes the behavior of isDeferred. If it's going to be deprecated it should have the same behavior until it is removed.

util/can.js
nlundquist: Yeah this works with jQuery deferreds.

util/can.js
nlundquist: Totally right.

matthewp:
@daffl should this go into major?

test/pluginified/2.0.5.test.js
```diff
@@ -316,7 +316,7 @@ var __m1 = (function () {
 				update: function () {
 					var deferred = this.scope.attr('deferreddata'),
 						scope = this.scope;
-					if (can.isDeferred(deferred)) {
+					if (can.isPromise(deferred)) {
```
justinbmeyer: no changes should be made in this file.  This is our legacy test.

PR #2268, opened 2016-02-15
Fix 'Object doesn't support this action' error in IE

### LINKED ISSUE #2227: console.log.apply not supported on IE11?
Hi guys!
   When using I'm getting a wired 'Object doesn't support this action' error in [can/util/can.js, `_logger` function](https://github.com/canjs/canjs/blob/master/util/can.js#L264)
   For me, the fix was to replace the `if-else` with a `try-catch`.

meconcomputer:
This should resolve #2227 issue

### DIFF
--- a/util/can.js
+++ b/util/can.js
@@ -261,9 +261,9 @@ steal(function () {
 			}
 		},
 		_logger: function (type, arr) {
-			if (console.log.apply) {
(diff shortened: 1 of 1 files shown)


daffl:
Sweet, thank you!

PR #2279, opened 2016-02-18
Omit unserialized route map properties from string coercion.

### LINKED ISSUE #2277: Properties on a can.route.map are coerced to strings when they should not be
I believe this issue is caused by this:

https://github.com/canjs/canjs/pull/2246/files

...and lives in isolation here:

https://jsfiddle.net/ryanwheale/0wyreatk/

nlundquist:
Resolves #2277

### DIFF
--- a/route/route.js
+++ b/route/route.js
@@ -121,8 +121,8 @@ steal('can/util', 'can/map', 'can/list','can/util/string/deparam', function (can
 		stringCoercingMapDecorator = function(map) {
 			var typeSuper = map.__type;
 			map.__type = function(val, prop) {
-				var serializable = this.define === undefined || this.define[prop] === undefined || !!this.define[prop].attr('serialize'), 
-					ret = typeSuper.apply(map, arguments); 
+				var serializable = this.define === undefined || this.define[prop] === undefined || !!this.define[prop].attr('serialize'),
+					ret = typeSuper.apply(map, arguments);
 				
 				if (serializable) {
 					return stringify(ret);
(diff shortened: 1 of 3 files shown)


route/route.js
```diff
@@ -120,8 +120,15 @@ steal('can/util', 'can/map', 'can/list','can/util/string/deparam', function (can
 		// change type coercion during Map setter to coerce all values to strings 
 		stringCoercingMapDecorator = function(map) {
 			var typeSuper = map.__type;
-			map.__type = function() {
-				return stringify(typeSuper.apply(map, arguments));
+			map.__type = function(val, prop) {
+				var serializable = this.define === undefined || this.define[prop] === undefined || !!this.define[prop].attr('serialize'),
```
dylanrtt: I think the last part should be `!!this.define[prop].serialize` since the define block should not be a map.

route/route_test.js
```diff
@@ -761,7 +761,31 @@ steal("can/route", "can/test", "steal-qunit", function () {
...
-				}, 30);
+				}, 5);
+			});
+		});
+
+		test("updating unserialized prop on bound can.Map causes single update without a coerced string value", function() {
+			expect(1);
+        
+			setupRouteTest(function (iframe, route) {
+				var appVM = new can.Map({define: {
```
dylanrtt: I think you want `can.Map.extend()` here.

dylanrtt:
Can we remove the `can.route.attr()` override now that the backing map handles the coercion? I don't think it's necessary, and it's currently not checking if the properties are serializable.

nlundquist:
Yes, but only with further changes. The new override to __type only coerces the values to strings. In can.route.attr keys are being string coerced, this would have to be added to a different backing map override.

nlundquist:
My last commit does this. Removes the .__type override & route.attr behaviour. Moves all string coercion to a .attr override. Resolved an issue I had when originally trying to do this in an .attr override.

PR #2282, opened 2016-02-22
Don’t include %root in a component’s initial viewModel data

### LINKED ISSUE #2250: `Type` in a component’s VM doesn’t work correctly with lists
If a `can.Component` has a `define` property that’s `Type` is a `can.List`, and that `can.List` has a `Map` type, and that `can.Map` has `define` properties, then those properties should be available on instances of the items within the list. For example:

```
var Item = can.Map.extend({
  define: {
    prop: {
      get: function() {
        return 'define prop';
      }
    }
  }
});

var Collection = can.List.extend({
  Map: Item
}, {});

var collectionInstance = new Collection([{}]);
collectionInstance[0].attr('prop');// Returns “define prop”

var homeTemplate = can.view.stache('<p>{{vmProp}}</p>');

can.Component.extend({
  tag: "home-page",
  template: homeTemplate,
  viewModel: {
    define: {
      collection: {
        Type: Collection
      },
      vmProp: {
        get: function() {
          var collection = this.attr('collection');
          return collection[0].attr('prop') || 'no prop';
        }
      }
    }
  }
});

var appTemplate = can.view.stache('<home-page collection="{collection}" />');

$("body").html(
  appTemplate({
    collection: [{objectProp:'object prop'}]
  })
);

// Body should have “define prop” but instead has “no prop”
```

Works in 2.2.9 (“define prop” is output): http://jsbin.com/cocarikefo/1/edit?html,js,output

Doesn’t work with latest (“no prop” is output): http://jsbin.com/rufavacera/1/edit?html,js,output

chasenlehara:
When a component is set up, `%root` was added to the initial `viewModel` data so it would be available in every scope. However, [`can.view.Scope` already has a special case for %root](https://github.com/canjs/canjs/blob/cee94ae7744ea6376e1f4bd7f6088b325db06555/view/scope/scope.js#L91-L94), so it’s not necessary for `can.Component` to do that.

Fixes #2250

### DIFF
--- a/component/component.js
+++ b/component/component.js
@@ -96,9 +96,7 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/view/bindings
 			setup: function (el, componentTagData) {
 
 				// Setup values passed to component
-				var initialViewModelData = {
-						"%root": componentTagData.scope.attr("%root")
-					},
+				var initialViewModelData = {},
 					component = this,
 					// If a template is not provided, we fall back to
 					// dynamic scoping regardless of settings.
--- a/component/component.js
+++ b/component/component.js
@@ -96,7 +96,9 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/view/bindings
 			setup: function (el, componentTagData) {
 
 				// Setup values passed to component
-				var initialViewModelData = {},
+				var initialViewModelData = {
+						"%root": componentTagData.scope.attr("%root")
+					},
 					component = this,
 					// If a template is not provided, we fall back to
 					// dynamic scoping regardless of settings.
--- a/component/component.js
+++ b/component/component.js
@@ -96,9 +96,7 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/view/bindings
 			setup: function (el, componentTagData) {
 
 				// Setup values passed to component
-				var initialViewModelData = {
-						"%root": componentTagData.scope.attr("%root")
-					},
+				var initialViewModelData = {},
 					component = this,
 					// If a template is not provided, we fall back to
 					// dynamic scoping regardless of settings.
@@ -130,6 +128,9 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/view/bindings
 				if(setupBindings) {
 					teardownFunctions.push(bindings.behaviors.viewModel(el, componentTagData, function(initialViewModelData){
 						
+						// Make %root available on the viewModel.
+						initialViewModelData["%root"] = componentTagData.scope.attr("%root");
+
 						// Create the component's viewModel.
 						var protoViewModel = component.scope || component.viewModel;
 						if (component.constructor.Map) {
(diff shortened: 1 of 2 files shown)


matthewp:
It's necessary for people that do `this.attr("%root")` though. Before can-wait we were telling people to do:

```js
var promise = SomeModel.getList();
this.attr("%root").waitFor(promise);
return promise;
```

Does this break something for you?

justinbmeyer:
Scopes have it, but not component viewModels. This needs to be present for ssr.

chasenlehara:
Thanks guys. Based on what I found, I think I could fix this by just moving that line of code. I’ll write a test for `%root` first, make sure this fix breaks it, then see if I can move that line of code to satisfy both tests.

@matthewp, yes, [this commit](https://github.com/canjs/canjs/commit/4590703c0ede61dd6a7da1130380116532a24981) is what first started breaking [my test case](https://github.com/canjs/canjs/commit/30a614218e42a19d95993452785689a9dec79e40) for #2250. I determined that when `%root` was being set up, it was not correctly initializing my scope variable.

chasenlehara:
@justinbmeyer & @matthewp, here’s what I’ve done:

- [Add a (really simple) test for `%root`in `can.Component`](https://github.com/canjs/canjs/commit/bc93096620920aa150de21938cffef5e1ad166ec)
- [Revert my previous changes](https://github.com/canjs/canjs/commit/1bbb2765cdf70413be0b0d260cab8d03f121bca2)
- [Fix the original bug by allowing the component’s attributes to be bound first before `%root`](https://github.com/canjs/canjs/commit/96d59b9cd5c9a65995108016a5a8db3ed5838f67)

component/component.js
```diff
@@ -130,6 +128,9 @@ steal("can/util", "can/view/callbacks","can/view/elements.js","can/view/bindings
 				if(setupBindings) {
 					teardownFunctions.push(bindings.behaviors.viewModel(el, componentTagData, function(initialViewModelData){
 						
+						// Make %root available on the viewModel.
+						initialViewModelData["%root"] = componentTagData.scope.attr("%root");
+
```
justinbmeyer: I'm not sure how this is working because it's not adding it to `initialViewModelData._data`

justinbmeyer:
@chasenlehara I'm curious why the initial data would cause the breaking behavior you identified.  It doesn't make sense to me, which makes me think there's a better fix somewhere.

Also, you should be able to demonstrate this w/o a component.  Just create the VM in the same way can.Component does, with initial data like `new YourMap({"%root": new can.Map()})`

chasenlehara:
The cause of this issue is a little tough to explain, but here’s my best shot:

- `can.Map` helpfully keeps track of objects that it’s converted, so if you pass it an object twice, it’ll return the same `can.Map` instance
- When `%root` is a plain object, it gets converted to a `can.Map`

In #2250, `%root` has a `collection` that is converted into a `can.List` because `%root` is converted first. Then, when `collection` is defined on the component, `can.Map` (unhelpfully in this case) returns the `can.List` it had already created instead of converting the object to the special `Collection` type.

This PR “fixes” (it’s kind of a hack) the issue by setting up `%root` after the other properties are converted, so it references the converted values instead of creating its own.

@justinbmeyer and I talked about this and he might be able to add more clarification to this rough description.

justinbmeyer:
@chasenlehara good explanation.

PR #2289, opened 2016-02-23
Upgrade to can-simple-dom 0.3.0

matthewp:
This is needed to release donejs 0.7.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 		"github": "https://github.com/canjs/canjs"
 	},
 	"dependencies": {
-		"can-simple-dom": "^0.3.0-pre.2",
+		"can-simple-dom": "^0.3.0",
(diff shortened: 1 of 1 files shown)


daffl:
I don't think this is necessary. `^0.3.0-pre.0` should match every `0.3.x` release (tested with http://semver.npmjs.com/)

matthewp:
That only happens on fresh npm installs though, right? I think of someone did an `npm update` they wouldn't be upgraded.

daffl:
I just tested `npm update` on my local CanJS. Before:

```
can-simple-dom@0.3.0-pre.3
```

After:

```
can-simple-dom@0.3.0
```

So it seems to work.

daffl:
We'll get the correct version range into the next release

PR #2298, opened 2016-03-01
Bugfix: elems must be an array for can.inserted()

lastzero:
Without this fix, `canjs/util/inserted/inserted.js` can fail at line 7...
```
if(!elems.length) {
  return;
}
```
...since length is undefined:
```
TypeError: Cannot read property 'length' of undefined
```


justinbmeyer:
@lastzero how did this fail?

justinbmeyer:
meaning ... is there some code that you ran that caused the error?

lastzero:
It was caused by some jquery plugin - probably select2. We came across this while refactoring existing code...

lastzero:
Not select2, but some datetime picker in conjunction with moment.js.

daffl:
That makes sense. One of those plugins probably called the method with `null` which seems to be valid in jQuery.

lastzero:
Yeah, I think arguments was [null]

PR #2319, opened 2016-03-09
Fix autorender template tag attributes. 

### LINKED ISSUE #2317: autorendered template attributes are not added to template viewModel
According to [the docs](https://canjs.com/docs/can%7Cview%7Cautorender.html) the following should work:

```
<script type="text/stache" can-autorender message="hello">
    {{message}}
</script>
```

However in the current (2.3.20) release the message attribute is not available to the template and the output is blank.

This is because an element (line 48) in can/view/autorender is jquery wrapped when it's not expected to be.

nlundquist:
Get attribute list from unwrapped element. Add tests.

Resolves #2317

### DIFF
--- a/view/autorender/autorender.js
+++ b/view/autorender/autorender.js
@@ -39,7 +39,7 @@ steal("can/util", "can/map/app", "can/util/view_model", function(can, AppState){
 		}
 	}
 	function setupScope(el) {
-		$el = can.$(el);
+		var $el = can.$(el);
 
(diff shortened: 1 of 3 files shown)


pYr0x:
this is fixed in can 3.0 but we should except that for 2.3.25

phillipskevin:
This PR is no longer needed as https://github.com/canjs/canjs/issues/2317 was closed.

PR #2330, opened 2016-03-17
Removing bower and non-jquery libraries

BigAB:
This PR removes the bower dependancy along with all the non-jquery base libraries (YUI, Dojo, mootools, zepto)

It is a step in the direction of a simplified **can-core** and a stand-alone can with no external dependancies.


daffl:
:shipit:

PR #2337, opened 2016-03-21
Remove all plugins

daffl:
This pull requests removes all non-core plugins for CanJS 3.0. Plugins have been migrated to their own repositories and modules in the CanJS organization.


justinbmeyer:
@daffl I was thinking super should be core.  What do you think?

daffl:
Should it not be a plugin at all then?

justinbmeyer:
What do you mean by plugin?  If you mean should it be in its own modlet, then yes it should.  Its still a good practice to have things broken out.

daffl:
Makes sense. I added it back in. All other plugins have been removed and now live in their own repository.

PR #2341, opened 2016-03-25
Fix the way we get PhantomJS 2 for the tests

BigAB:
See https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-200965782 for more details


BigAB:
Alright, that didn't work.

Back to the drawing board.

justinbmeyer:
Shouldn't we avoid phantom and use FF instead?

PR #2344, opened 2016-03-28
Use YouTube embed link to avoid cross origin errors

Dan-Nolan:
Getting domain issues with a couple of the YouTube videos in the docs, so I can't view them directly:

canjs_-_can_fixture

The rest use the embed link and work great. They are very helpful!


marshallswain:
Great catch!  Thanks for taking the time to fix it.

PR #2355, opened 2016-04-01
Standalone CanJS

daffl:
Work in progress

This pull request removes the dependencies on separate libraries making CanJS usable standalone in any modern browser.


daffl:
Note: https://github.com/kenwheeler/cash might be a good replacement for the jQuery event system (full library might potentially even allow to migrate existing jQuery apps).

justinbmeyer:
Is there something difficult with replacing the event system?  My hope is that IE9 would be pretty standards and no huge hurdles to work around.

PR #2357, opened 2016-04-04
Add exclusive mapping to map/define

RALifeCoach:
When using map.define I added an option to specify ```exclusiveMapping: true```. If set the code will only use the attributes defined in the map when building the local copy of the object. Useful when the server based object has hundreds of properties, but the UI only uses a few of them.


pYr0x:
we need a test to except that PR. Also documentation is needed.
also there are some JSLint errors.

this feature maybe can released for the legacy canjs 2.3

pls let me know if you want to resolve the problems

pYr0x:
so i will close this

PR #2366, opened 2016-04-08
Make can-define work with can.Component

akagomez:
- Enable a non-can.Map constructor to be set as the `viewModel`
- Observe object reads
- Enable compute reads without `.attr()`


pYr0x:
can we close this?

PR #2390, opened 2016-05-10
Fix mustache core mapping

### LINKED ISSUE #2389: mustache_core not included in browserify mappings
https://github.com/canjs/canjs/blob/master/Gruntfile.js

ccummings:
Adds a test for using a compiled stache template with Browserify.

Fix is to add a `browser` mapping for `view/stache/mustache_core`

Closes #2389


matthewp:
:(

Issue #2398, opened 2016-05-19
Sort plugin is activated via .sort()


akagomez:
See this bin: http://jsbin.com/zabibun/2/edit?html,console,output

Essentially, after the changes made in https://github.com/canjs/canjs/pull/2257 there's no way for somebody to sort a list without the sort plugin setting up bindings.

The solution is to separate the two features via the api (comparator to bind, `sort()` for on-demand sorting) and provide an alternative solution for this issue: https://github.com/canjs/canjs/issues/2159

justinbmeyer:
Why is there no way to sort a list without setting up bindings?  What specifically enforces that?

akagomez:
https://github.com/canjs/canjs/blob/master/list/sort/sort.js#L236

I suppose that's not entirely true. You can sort a list without setting up bindings if you don't provide a comparator.

justinbmeyer:
Why couldn't it do:

```
this.attr('comparator', comparator);
this._sort();
```

?

akagomez:
Closed via #2399.

PR #2399, opened 2016-05-19
Pass list items to .sort(fn) and don't set the comparator

### LINKED ISSUE #2159: sort plugin forces comparing 'comparator' values, even when overriding the comparator function
I use a custom sort comparator function on a can.Map.List. In 2.3.x (I guess with the changes to the sort plugin) this breaks down. In 2.2.x it works.

When I remove the default comparator string on my model, the sorting works again. 
What's happening is that instead of the item itself, the value specified by the original comparator is used instead.

Culprits seem to be:
https://github.com/canjs/canjs/blob/master/list/sort/sort.js#L207-L209
and
https://github.com/canjs/canjs/blob/master/list/sort/sort.js#L251-252

I hope this behaviour isn't intentional.

akagomez:
This PR does two things: 

- `list.comparator('prop'); list.sort(fn)` - `fn` will be passed the list items and **not** the item's `prop` value (#2159)
- `list.sort(fn)` - `fn` will **not** be set as `list.attr('comparator', fn)` under the hood thus preserving the ability to call `.sort()` at will without setting up sorting for every "change" event (#2398)

Closes #2159, and #2398.

### DIFF
--- a/list/sort/sort.js
+++ b/list/sort/sort.js
@@ -200,13 +200,10 @@ steal('can/util', 'can/list', function (can) {
 		/**
 		 * @returns {number} The value that should be passed to the comparator
 		 **/
-		_getComparatorValue: function (item, overwrittenComparator) {
+		_getComparatorValue: function (item, singleUseComparator) {
 
 			// Use the value passed to .sort() as the comparator value
-			// if it is a string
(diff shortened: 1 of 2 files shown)


pYr0x:
@akagomez @daffl we should also make a PR against https://github.com/canjs/can-list-sort ?

PR #2407, opened 2016-05-27
Fix incorrect reference to `bind` in the `delegate` documentation.

chrischrischris:
The example will not work with the bind syntax.


chrischrischris:
There are additional errors with the documentation, will get all of the fixes in one pull request.

Issue #2418, opened 2016-06-15
SVG camelCased tags are lowercased by Stache


marshallswain:
I have an SVG file that partially disappears after loaded with Stache.  With SSR enabled, the initial load of the site shows the full image as it should, but upon re-attachment of the app, the center disappears again.  I turned on network throttling to see if I could capture the difference and discovered that camelCased SVG tags like `radialGradient` are getting lowercased to `radialgradient`, which is breaking the rendering of the image.

Here's the SVG as loaded from the SSR server:
```svg
<svg class="protect" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns" width="30px" height="37px" viewBox="0 0 30 37" version="1.1">
    <defs>
        <radialGradient cx="49.995417%" cy="49.9971135%" fx="49.995417%" fy="49.9971135%" r="57.036852%" id="radialGradient-1">
            <stop stop-color="#F9F9F9" offset="0%"></stop>
            <stop stop-color="#F9F4CA" offset="100%"></stop>
        </radialGradient>
    </defs>
    <g id="Account" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
        <g id="Short-Header" sketch:type="MSArtboardGroup" transform="translate(-500.000000, -11.000000)">
            <g id="Header" sketch:type="MSLayerGroup" transform="translate(20.000000, 10.000000)">
                <g id="Protect" transform="translate(480.000000, 1.000000)" sketch:type="MSShapeGroup">
                    <g id="protect-icon">
                        <path d="M15,0 L0,3.55307369 L0,25.5425735 C0,28.0471827 6.035,31.6486936 6.035,31.6486936 L15,37 L23.965,31.6486936 C23.965,31.6486936 30,28.0471827 30,25.5425735 L30,3.55307369 L15,0 L15,0 Z" id="Shape" fill="#C1272D"></path>
[...]

PR #2420, opened 2016-06-16
Support svg camelCase elements.

marshallswain:
This adds another map for all of the svg tags that require camelCasing. Taken from this page: https://developer.mozilla.org/en-US/docs/Web/SVG/Element

Also includes a passing test for radialGradient.

related to canjs/canjs#2418


marshallswain:
This same fix has already been merged into the v3.0 repo here: https://github.com/canjs/can-view-parser/pull/9

pYr0x:
so we can set this for the legacy release of 2.3.X ?

justinbmeyer:
@marshallswain why do we have to even track the `caseMatters`?  Could we just stop `toLowerCase()`-ing what we get?

marshallswain:
That makes sense to me. I don't know why we are lower casing in the first
place.

PR #2428, opened 2016-06-30
Fix computedAttr.compute is not a function

wongpratan:
http://forums.donejs.com/t/computedattr-compute-is-not-a-function-after-upgrade/204


justinbmeyer:
Thanks @wongpratan.  @chasenlehara why the ready label when there's no test?

matthewp:
I think that label is automatically added by waffle any time a PR is submitted.

PR #2436, opened 2016-07-12
Set readOnly property based on val

### LINKED ISSUE #2431: Readonly attr is always true
Example:
http://jsbin.com/nejayewaqu/edit?js,console,output

Suspected cause:
https://github.com/canjs/canjs/blob/a3b566ceb93dc60e4f1e909f9a27123b4ff55771/util/attr/attr.js#L65-L68

akagomez:
Closes #2431.

### DIFF
--- a/util/attr/attr.js
+++ b/util/attr/attr.js
@@ -63,7 +63,7 @@ steal("can/util/can.js", function (can) {
 				"checked": true,
 				"disabled": true,
 				"readonly": function (el, val) {
-					el.readOnly = true;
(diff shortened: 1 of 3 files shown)


pYr0x:
thx @akagomez , but pls make also a PR against canjs/can-util (https://github.com/canjs/can-util/blob/0ff04106cfd5170727e6bc2f6c0041443e9f481d/dom/attr/attr.js)

and pls dont forget this issue you fixed in can-list-sort to implement into canjs/can-list-sorthttps://github.com/canjs/can-list-sort/issues/3

PR #2438, opened 2016-07-16
fix svg-use not displaying svg-image (#2384)

colin-tucows:
fix svg-use not displaying svg-image (i.e. element is collapsed) because namespace-specific 'el.setAttributeNS()' must be used to set 'xlink:href'-attribute instead of 'el.setAttribute() (#2384)

### DIFF
--- a/util/attr/attr.js
+++ b/util/attr/attr.js
@@ -192,6 +192,11 @@ steal("can/util/can.js", function (can) {
 								el.setAttributeNode(node);
 							} else {
 								el.setAttribute(attrName, val);
+
+								// fix svg-use not displaying svg-image (i.e. element is collapsed) because namespace-specific 'el.setAttributeNS()' must be used to set 'xlink:href'-attribute instead of 'el.setAttribute()' (#2384)
+								if(el.tagName === "use" && attrName === "xlink:href"){
+									el.setAttributeNS("http://www.w3.org/1999/xlink", "href", val);
+								}
 							}
 						};
 					}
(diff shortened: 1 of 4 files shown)


pYr0x:
Thank you. Looks good. But the test only covers that the namespace is set right, but is the image really shown?
Can you also adapt Canjs 3.0?

colin-tucows:
@pYr0x : checking the svg-element and/or use-element clientWidth/Height were not sufficient checks, so i will have to investigate further on confirm the use-element was able to retrieve the svg-image/file.

 i am busy for the next few months with a launch. i will try to get back to this when i can.

util/attr/attr.js
```diff
@@ -192,6 +192,11 @@ steal("can/util/can.js", function (can) {
 								el.setAttributeNode(node);
 							} else {
 								el.setAttribute(attrName, val);
+
+								// fix svg-use not displaying svg-image (i.e. element is collapsed) because namespace-specific 'el.setAttributeNS()' must be used to set 'xlink:href'-attribute instead of 'el.setAttribute()' (#2384)
+								if(el.tagName === "use" && attrName === "xlink:href"){
```
justinbmeyer: are there other attributes that should be set this way?  Should anything with `:` be set this way?

phillipskevin:
@colin-tucows This was fixed in 3.0 to use `el.setAttributeNS(...)` for any attribute containing a `:`. Can you take a look at the fix [here](https://github.com/canjs/can-util/pull/98/files) and apply the same solution? Also, the pull request will need to be made to the `2.3-legacy` branch since master now has the 3.0 code.

Issue #2444, opened 2016-08-02
can-autorender does not attach viewModel to the element.


ilyavf:
JSBin: [here](http://jsbin.com/repiko/edit?html,js,console,output)

Problem is [here](https://github.com/canjs/canjs/blob/master/view/autorender/autorender.js#L45): 
```
var scope = (can.data(el, "scope") || can.data(el, "viewModel")) ?
			can.viewModel(el) : new AppState();
```
`can.viewModel(el)` creates a new VM and attaches it to the el. But in this case it does not get called thus VM=`new AppState()` is not attached to the el.

ilyavf:
This PR also fixes #2319 Fix autorender template tag attributes.

PR #2445, opened 2016-08-02
#2444 autorender did not attach vm to the element

ilyavf:
Attach VM to the autorender element when creating a new one.

### DIFF
--- a/view/autorender/autorender.js
+++ b/view/autorender/autorender.js
@@ -1,5 +1,5 @@
 "format steal";
-steal("can/util", "can/util/view_model", function(can){
+steal("can/util", "can/map/app", "can/util/view_model", function(can){
 	
 	var deferred = new can.Deferred(),
 		ignoreAttributesRegExp = /^(dataViewId|class|id|type|src)$/i;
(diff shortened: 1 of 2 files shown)


matthewp:
Can you add a test?

matthewp:
can/map/app doesn't have any use any more and should be removed. Use can/map here instead.  In 3.0 it will need to be types.DefaultMap.

ilyavf:
@matthewp I removed can/map/app usage, so its back to pre-[39c08a](https://github.com/canjs/canjs/commit/39c08a421c21f7343fa41c88fdb7caaaae141e31) state:
```
// This will create a new scope (if it does not exist) and attach it to `el` with data:
var scope = can.viewModel(el);
```
Tests are there.

matthewp:
Failing in Travis, is it working locally?

ilyavf:
@matthewp Locally I am getting stuck on amd tests (4159 - on the same test as Travis)
```
Running "testee:steal" (testee) task
....... 
22850 passing (4m)
Running "testee:amd" (testee) task
... 4159 dots ... and stuck
```

ilyavf:
@matthewp: Woohoo, the CI is passing. So, I had to import `can/map/app` from inside autorender.js. Do you have an idea why?

matthewp:
No, can/map/app should be removed actually. So there must be something wrong here. Which test is failing without it?

matthewp:
We can fix can/map/app in another issue, or not at all since can-view-autorender doesn't depend on it.

PR #2512, opened 2016-09-21
Add legacy can/construct/super steal target

bmomberger-bitovi:


justinbmeyer:
@airhadoken Thanks!  I don't think this is exactly legacy.  I'd say more "ecosystem".  Will this show up under "ecosystem"?

bmomberger-bitovi:
It's referenced as a plugin by can-construct in the docs.  I'm terming this target "legacy" not because can-construct-super is legacy but because its support is being backported to the legacy canjs monolith.

PR #2524, opened 2016-09-27
Enable IE9 on SauceLabs

daffl:
The TodoMVC tests are passing locally on my test VM but fail on Saucelabs.


daffl:
Missing browsers will be reopened against master

PR #2532, opened 2016-09-28
Allow setters to return undefined

### LINKED ISSUE #2523: Setting undefined to a map property with a setter won't remove previously set value
```javascript
var MyMap = can.Map.extend({
  define: {
    foo: {
      set: function(newVal) {
        return newVal;
      }
    }
  }
});

var map = new MyMap();

map.attr('foo', 'bar');
assert(map.attr('foo') === 'bar', 'foo should be bar');

map.attr('foo', undefined);
assert(typeof map.attr('foo') === 'undefined', 'foo should be undefined');  
// ^ This fails because map.attr('foo') returns 'bar'
```

JSBin with code http://jsbin.com/nakiwurivo/1/edit?js,console

matthewp:
Fixes #2523

The previous behavior of ignoring returning undefined should only apply
to asynchronous setters, so checking for the correct argument length to
handle that scenario and otherwise allowing setting a property to
undefined.

### DIFF
--- a/map/define/define.js
+++ b/map/define/define.js
@@ -169,7 +169,7 @@ steal('can/util','can/map/map_helpers.js', 'can/map', 'can/compute', function (c
 					return;
 				}
 				// if it took a setter and returned nothing, don't set the value
-				else if (setValue === undefined && !setterCalled && setter.length >= 1) {
+				else if (setValue === undefined && !setterCalled && setter.length > 1) {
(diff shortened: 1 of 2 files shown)


matthewp:
@daffl @justinbmeyer is there a way to tag a PR to go in the next 2.3.x (if that even happens)?

PR #2554, opened 2016-10-11
Adding Opera on Linux to Saucelabs

daffl:


daffl:
All passing locally. We'll use Safari to test against Webkit.

PR #2601, opened 2016-10-20
Added can-stream to the package.json

obaidott:
Added can-stream to the package.json, so it can be included in the documentation.


daffl:
Can you rebase with the latest master?

Issue #2745, opened 2016-11-14
Travis job can timeout before iOS simulator launches


phillipskevin:
Travis jobs will timeout after 10 minutes if nothing is output to the console. Sometimes it takes longer than that to launch an iOS simulator (like [here](https://travis-ci.org/canjs/canjs/builds/175760258) and [here](https://travis-ci.org/canjs/canjs/builds/175760143)).

This can be fixed using [travis_wait](https://github.com/canjs/canjs/pull/2744/commits/a530f05c9a6baed821f8c70b43c6c509c9754624), but then the output of `npm run ci` is not seen until after the entire job completes, which makes it really difficult to tell what is going on.

To fix this, we should print something to stdout while waiting for the saucelabs job after calling [`driver.init`](https://github.com/canjs/canjs/blob/master/test/test-sauce-labs.js#L127) and until [the job is submitted](https://github.com/canjs/canjs/blob/master/test/test-sauce-labs.js#L133).

PR #2746, opened 2016-11-14
Preventing Travis from timing out while Sauce Labs is initialized

phillipskevin:
Closes https://github.com/canjs/canjs/issues/2745.


phillipskevin:
CI failed because of the XHR Shim test which will be fixed by https://github.com/canjs/canjs/pull/2744.

PR #2749, opened 2016-11-15
Make sure node modules folder is not ignored

### LINKED ISSUE #2740: Demos aren't working
On http://v3.canjs.com/doc/can-component.html and everywhere else, the demo pages aren't working:

canjs_-_can-component

This is because steal.js is a 404.  

canjs_-_can-component

However, it is in gh-pages:

https://github.com/canjs/canjs/blob/gh-pages/node_modules/steal/steal.js

The `Makefile` is certainly adding it:

https://github.com/canjs/canjs/blob/master/Makefile#L10

m-mujica:
Github pages was recently moved to jekyll 3.3 which ignores "node_modules" and "vendors" directories, adding an empty ".nojekyll" file will prevent those directories to be ignored.

See https://www.bennadel.com/blog/3181-including-node-modules-and-vendors-folders-in-your-github-pages-site.htm

Closes #2740


phillipskevin:
👍 once tests pass.

Issue #2753, opened 2016-11-16
Clean up "who's using" page


moschel:
Use better screenshots for this page: http://v3.canjs.com/doc/guides/who-uses-canjs.html

tomgreever:
@brianmoschel what do you think about something like this? This adds a border and dropshadow, which makes them look nicer. And for the ones that include a url, we can fake a browser address bar. This won't change the markdown other than wrapping the section in a parent div. Other than that it's just CSS, so it should still be simple to update.

![screen shot 2016-11-16 at 12 33 48 pm](https://cloud.githubusercontent.com/assets/326843/20360351/f9fdb17e-abf8-11e6-9c62-a2545ed4bbec.png)

tomgreever:
oh and it scales nicely. 
![screen shot 2016-11-16 at 12 36 46 pm](https://cloud.githubusercontent.com/assets/326843/20360439/6222f55c-abf9-11e6-9774-6cfdad304086.png)

moschel:
looks good to me

tomgreever:
This should be closed by #2757 - I'm not seeing the changes on GH pages yet, but closing anyway and trusting it's gtg since the PR was merged

PR #2757, opened 2016-11-16
New screenshots

tomgreever:
Adds better screenshots for https://github.com/canjs/canjs/issues/2753 and relies on https://github.com/canjs/bit-docs-html-canjs/pull/194

When complete, screenshots should look like this. But I don't have a good way of testing this locally, so I'm really not sure what it will do! 

screen shot 2016-11-16 at 4 12 19 pm


tomgreever:
@brianmoschel or @justinbmeyer can you review this and https://github.com/canjs/bit-docs-html-canjs/pull/194 and let me know when it's all merged and visible? I don't have a good way of checking my changes from bit-docs in this repo, so I'm just :fingers_crossed: that it's all good.

Issue #2794, opened 2016-11-29
Having issues using Browserify with CanJS 3


chasenlehara:
When I tried to use a module like `can-define/map/map`, I get this error when I build with Browserify:
```
Error: Cannot find module 'cssify' from '~/canjs-browserify-npm/node_modules/can-observation'
    at ~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at ~/.nvm/versions/node/v6.9.1/lib/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:123:15)
```

I’ve created a repo to demo this issue. Here’s the code in my `can-define-map` branch that demonstrates the issue: https://github.com/chasenlehara/canjs-browserify-npm/blob/can-define-map/main.js

Requiring a module like `can-util` works correctly (`master` branch): https://github.com/chasenlehara/canjs-browserify-npm/blob/master/main.js

Please let me know if this is a CanJS bug or if I need to do more setup to make it work with Browserify.

matthewp:
The problem seems to be that a bunch of packages have a `cssify` transform listed even though they don't have anything to do with CSS. Going to post subissues as I find them:

https://github.com/canjs/can-construct/issues/19

matthewp:
Those are the only issues.

matthewp:
Fixed those two issues and it looks to be working fine now. Reopen if you see otherwise.

chasenlehara:
Reopening because `can-stache` has the same issue: https://github.com/canjs/can-stache/blob/master/package.json#L31

matthewp:
A convenient link: https://github.com/search?p=1&q=org%3Acanjs+cssify&type=Code

matthewp:
@chasenlehara this should be fixed now.

chasenlehara:
[...]

PR #2812, opened 2016-11-29
Work In Progress: Update the “Setting up CanJS” guide

### LINKED ISSUE #2642: Setting Up CanJS
- how to make a custom build.

chasenlehara:
- Remove references to the zip download, the download builder, and Bower
- Update the JS Bin code
- Remove mentions and uses of jQuery because it isn’t required
- Update the examples to use good CanJS 3 code
- Remove references to DoneJS because it doesn’t support CanJS 3 yet
- Replace the AMD and RequireJS section with a note that they’re no longer supported
- Remove references to `can-compile` because it’s not supported

Closes #2642

### DIFF
--- a/docs/can-guides/experiment/setting-up-canjs.md
+++ b/docs/can-guides/experiment/setting-up-canjs.md
@@ -2,93 +2,36 @@
 @parent guides/experiment 5
 @outline 2
 
-@description CanJS is packaged in multiple ways so that it can fit into any development workflow. Learn how to setup CanJS in different environments.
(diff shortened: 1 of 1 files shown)


chasenlehara:
I still have a few things I’d like to do:
- [ ] Verify the Browserify code works after [this issue](https://github.com/canjs/canjs/issues/2794) is fixed
- [ ] Add instructions for using Webpack
- [ ] Add a note about whether you can use Yarn

PR #2839, opened 2016-12-01
Add can-connect-feathers

marshallswain:


marshallswain:
Docs are here: https://github.com/canjs/can-connect-feathers/tree/2.0/_docs

PR #2908, opened 2016-12-19
Improve the migration guide

### LINKED ISSUE #2789: The migration guide doesn't mention steal-stache
Steal users are used to stache "just working" and there's no indication that something special needs to be done when upgrading to 3.0. http://v3.canjs.com/doc/migrate-3.html

chasenlehara:
- Combine the first two sections so the guide is a little easier to follow
- Add a section about using `steal-stache` (fixes #2789)
- Add a mapping between 2.3 paths and the new modules
- Fix various grammar/spelling/etc. issues


matthewp:
Looks good!

PR #2921, opened 2016-12-20
Improve the Migration Guide’s structure

### LINKED ISSUE #2920: Improve the Migration Guide’s structure
Here are the current sections in the [Migration Guide](http://canjs.com/doc/migrate-3.html):

1. Minimal migration path
	- Use can/legacy
	- Set leakScope on Components
	- Asynchronous inserted/removed events
	- can.view is no more
	- Promises are needed
	- Passing the this value to a helper
	- No more global scope names
	- Using stache templates with Steal
2. Modernized migration path
	- Use can-* packages
	- Wrap elements in jQuery objects
	- Remove use of change events
3. Future-proof migration path
	- can-define replaces can-map
	- Use can-connect directly
	- Use the new binding syntax

Following the last step isn’t really “future-proofing” your application, some of the sections are a little out of place, and I want to [add a section about what you can do _before_ upgrading](https://github.com/canjs/canjs/issues/2460), so here’s a proposal for how the guide should be organized:

1. Pre-migration preparation
	- Use the module folders
	- Replace uses of `can.$`
	- Set `leakScope` on components
2. Minimal migration path
	- Use `can/legacy`
	- Use `can-jquery/legacy`
	- Asynchronous `inserted` and `removed` events
	- Replace uses of `can.view`
	- Use real Promises
	- Passing the `this` value to a helper
	- No more global scope names
	- Use `steal-stache` for templates
2. Modernized migration path
	- Use the new can-* packages
	- Move from `can-map` to `can-define`
		- Remove use of change events
	- Use `can-connect` directly
3. Avoid future deprecations and removals
	- Wrap elements in jQuery objects
	- Use the new binding syntax

Thoughts/ideas/feedback?

chasenlehara:
Fixes #2920

### DIFF
--- a/docs/can-guides/commitment/migrating_to_3.0.md
+++ b/docs/can-guides/commitment/migrating_to_3.0.md
@@ -1,6 +1,7 @@
 @page migrate-3 Migrating to CanJS 3
 @parent guides/commitment 2
 @templateRender <% %>
+@description This guide walks you through the step-by-step process to upgrade a 2.x app to CanJS 3.
 @outline 0
 
 @body
--- a/docs/can-guides/commitment/migrating_to_3.0.md
+++ b/docs/can-guides/commitment/migrating_to_3.0.md
@@ -1,6 +1,7 @@
 @page migrate-3 Migrating to CanJS 3
 @parent guides/commitment 2
 @templateRender <% %>
+@outline 0
 
 @body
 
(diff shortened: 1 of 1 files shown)


docs/can-guides/commitment/migrating_to_3.0.md
```diff
@@ -1,54 +1,23 @@
 @page migrate-3 Migrating to CanJS 3
 @parent guides/commitment 2
 @templateRender <% %>
+@outline 0
```
justinbmeyer: no outline?  Why?

docs/can-guides/commitment/migrating_to_3.0.md
```diff
@@ -1,54 +1,23 @@
...
-  "dependencies": {
-    "can": "^<%canjs.package.version%>"
-  }
-}
-```
-
-The `^` ensures you get minor and patch releases as those are released.
-
-At a minimum, to upgrade your code for CanJS 3, you must make all of the following changes to your code:
+## Pre-migration preparation
```
justinbmeyer: You might want to redefine what this means here.

docs/can-guides/commitment/migrating_to_3.0.md
justinbmeyer: My guess is that it's trying to provide its own.  Also, there's no description (which show up if you hover over a link to this page).  I'm not sure it's worth breaking this convention.

chasenlehara: I’ll add the description. I thought the outline looked out of place because it’s summarized in the text below.

docs/can-guides/commitment/migrating_to_3.0.md
matthewp: I would just say **Preparation**

PR #2923, opened 2016-12-21
Use a specific snapshot for all of our JS Bins

chasenlehara:
Some of the links to JS Bin were for the latest version of the code, not a snapshot, so if the code changed, other users would see those changes.


justinbmeyer:
@chasenlehara if they aren't using `justinbmeyer` JSBins, will they work via https?  Does CanJS work via HTTPS?

chasenlehara:
The URLs that still have `http` in them are just links. I [already fixed](https://github.com/canjs/canjs/pull/2848) the embeds to be protocol relative.

justinbmeyer:
@chasenlehara I think only a paid account can be embedded in https.

chasenlehara:
They stopped making that a paid feature a year and a half ago: https://jsbin.com/blog/ssl-for-all

PR #2926, opened 2016-12-21
Skip modules using steal-clone in the production tests

marshallswain:


matthewp:
Do you mean the steal-clone module?

matthewp:
Rather than not running these tests, would it be better to just not run the *specific* tests? For example, only one can-observation test uses steal-clone: https://github.com/canjs/can-observation/blob/master/can-observation_test.js#L385

cc @phillipskevin

phillipskevin:
No, it is just importing `steal-clone` that causes errors. This is the issue https://github.com/stealjs/steal/issues/952.

PR #2964, opened 2017-01-11
WIP: Add can-model to the Legacy library list

### LINKED ISSUE #2793: Add can-model to the Legacy library list
The [Legacy](https://v3.canjs.com/doc/can-legacy.html) section should include [can-model](https://github.com/canjs/can-model/) so links to it (such as on the [migration](https://v3.canjs.com/doc/migrate-3.html) page) work correctly.

chasenlehara:
Five tests are failing when I run `npm run testee` locally.

I’m not sure why the `can-model` tests would work fine locally but not when included in this repo, so if someone has any guidance, it’d be much appreciated.

Fixes #2793


phillipskevin:
@chasenlehara looks like this is only failing after the merge. Did you try merging `2793-add-can-model-as-legacy` into `3-4` and then running the tests locally?

chasenlehara:
@phillipskevin, I know this isn’t the same as what you asked, but I merged `master` into this branch and am still seeing the same test failures.

chasenlehara:
…and just realized I should’ve changed the base.

chasenlehara:
Actually, on second thought, would this go into the next minor release? Should I update `minor` to match current `master` and then change the base of this PR to that?

PR #2978, opened 2017-01-23
Add can-validate repos for documentation and reference. Create demo.

Macrofig:
Adding can-validate, can-validate-validatejs, and can-define-validate-validatejs to the repo for documentation build.

Added demo for can-define-validate-validatejs.


phillipskevin:
@Macrofig It looks like it doesn't like `""can-define-validate-validatejs": "0.0.0"'. Can you take a look?

demos/can-validate/credit-card.html
```diff
@@ -0,0 +1,98 @@
...
+		var defineValidator = require("can-define-validate-validatejs");
+		var formatErrors = require("can-validate").formatErrors;
+		require("can-stache-converters");
+
+		stache.registerHelper('formatErrors', function (errors, format) {
+			errors = typeof errors === 'function' ? errors() : errors;
+			return formatErrors(errors, format);
+		});
+
+		var CreditCard = DefineMap({
```
justinbmeyer: DefineMap.extend

demos/can-validate/credit-card.html
```diff
@@ -0,0 +1,98 @@
...
+		label small.has-errors {
+			color: tomato;
+		}
+	</style>
+</head>
+<body>
+	<script type="text/stache" id="demo-html">
+		<form>
+			<div>
+				<label>Credit Card Number <small class="has-errors">{{formatErrors(testSet('card', card), 'object').card.join(', ')}}</small></label>
```
justinbmeyer: should we make `formatErrors` a bit nicer so it handles our single use case:

```html
{{formatErrors(this, "card"}}
```

Why is it `testSet`?  Presumably the value is already set as there is a two-way binding with `card` and not `{^$value}="*cardValue"`.

phillipskevin: @Macrofig any chance you can update this example? Or at least confirm that it could be done the way @justinbmeyer is proposing?

PR #3004, opened 2017-02-08
Version the install commands in the Migration guide

chasenlehara:
If you run `npm install can --save` in a 2.3 project, it’ll just install the latest version of 2.3, not CanJS 3.

### DIFF
--- a/docs/can-guides/commitment/migrating_to_3.0.md
+++ b/docs/can-guides/commitment/migrating_to_3.0.md
@@ -129,7 +129,7 @@ Component.extend({
 If you are already using `can` through npm, simply run the following command to install the latest version:
 
 ```
-npm install can --save
+npm install can@3 --save
 ```
 
 This will update your `package.json` to look something like this:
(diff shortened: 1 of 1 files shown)


justinbmeyer:
If you install canjs, but you should install can.

docs/can-guides/commitment/migrating_to_3.0.md
```diff
@@ -244,7 +244,7 @@ Instead, the first argument to [can-construct.extend] is the name of the constru
 If you use StealJS, you’ll need to install [steal-stache] to load your templates:
 
 ```
-npm install steal-stache --save
+npm install steal-stache@3 --save
```
justinbmeyer: was there another version of this?  Why would someone need to do this?

matthewp: maybe if there becomes a can 4.0 and you still want the migrating to 3.0 guide to work.

chasenlehara: Yeah, this is not required right now, but I think we should typically version our install commands so new and old guides install the correct version.

chasenlehara:
> If you install canjs, but you should install can.

Yes, sorry I mistyped my original description. It’s updated now.

PR #3027, opened 2017-02-28
change name to id

Rlin415:
The "name" attribute is deprecated. It was causing this issue: https://github.com/canjs/bit-docs-html-canjs/issues/208.

### DIFF
--- a/docs/can-guides/commitment/migrating_to_3.0.md
+++ b/docs/can-guides/commitment/migrating_to_3.0.md
@@ -323,7 +323,7 @@ Here’s a list of all the paths in CanJS 2.3 that now have separate modules in 
 - `can/view/stache/stache` — [can-stache]
 - `can/view/target/target` — [can-view-target]
 
-<a name="Future_proofmigrationpath"></a>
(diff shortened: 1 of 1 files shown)


matthewp:
I don't doubt you, but could you like to anything about name being deprecated? I wasn't aware of that.

Rlin415:
Sure! Here is one: https://www.w3schools.com/tags/att_a_name.asp

Rlin415:
It's also stated here as well: https://www.w3.org/TR/html5/obsolete.html

chasenlehara:
Heh, I missed the memo that `name` was deprecated too.

Here’s our code that doesn’t support `name`: https://github.com/canjs/bit-docs-html-canjs/blob/d1352069818e053453a5f082fad68e08f79f3ad6/static/canjs.js#L98

chasenlehara:
This document will need to be updated to not use `name`: https://github.com/donejs/donejs/pull/829/files#diff-54741d26fa44ca37b832eee981ddc15b

PR #3028, opened 2017-02-28
Do Not Merge: Upgrade to steal@1.x

marshallswain:
The tests are currently broken due to https://github.com/stealjs/steal/issues/1079


christopherjbaker:
fixed in #3381

PR #3073, opened 2017-03-09
Upgrade bit-docs-generate-html

Rlin415:
Npm wasn't upgrading bit-docs-generate-html to its latest minor version for some reason. I know this because I was still getting the console.logs that I removed in bit-docs/bit-docs-generate-html#14 even after I published a new version with the change.


chasenlehara:
Remember that when the major version is 0, `^` will only include new patch versions, not minor.

I think Travis is failing because we need to set the FIrefox version; I’ve opened a PR for that: https://github.com/canjs/canjs/pull/3074

chasenlehara:
@Rlin415 If you get a chance, could you please merge `master` into your branch and push? Hopefully that’ll fix the CI issue.

chasenlehara:
@Rlin415 It looks like a commit from https://github.com/canjs/canjs/pull/3097 was missed when this branch was updated. I don’t think it’ll cause any issues but it might be good to update it again.

PR #3075, opened 2017-03-09
Documentation for replacing can.construct.proxy

### LINKED ISSUE #2511: Migration Guide: Add can/construct/proxy replacement
The old functionality of `can/construct/proxy` is going away in 3.0 in favor of `.bind`. We should include a note about how to use .bind to get the same functionality.

Related: #2510

Rlin415:
This closes #2511


Rlin415:
Hmm, I guess you're right. The user could always read the documentation on `Function.prototype.bind` if they didn't already know how to use it.

chasenlehara:
@Rlin415 There’s a merge conflict that need to be fixed.

PR #3132, opened 2017-03-28
Finalize the Weather Report Guide

### LINKED ISSUE #3121: Recipe: Weather Report
Add a guide for creating a weather report app.

Open PR: https://github.com/canjs/canjs/pull/3061

chasenlehara:
- Change the starter JS Bin to not include the plain HTML so they don’t have to remove it at the beginning of the guide
- Add a Result section at the end
- Pin the JS Bins to specific versions instead of their latest versions
- Make the spacing consistent in the code examples
- Minor spelling & grammar fixes

Closes #3121

### DIFF
--- a/docs/can-guides/commitment/recipes/weather-report/weather-report.md
+++ b/docs/can-guides/commitment/recipes/weather-report/weather-report.md
@@ -9,19 +9,19 @@ CanJS 3.5.
 
 The final widget looks like:
 
-<a class="jsbin-embed" href="https://jsbin.com/vujugel/1/embed?js,output">JS Bin on jsbin.com</a>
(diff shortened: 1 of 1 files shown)


justinbmeyer:
The starter bin should include some example HTML.

phillipskevin:
When I presented this guide at DoneJS Chicago, a few people were confused about the starter JSBin having all of the HTML. Especially since the solution to the first problem is just a div containing `WEATHER WIDGET`.

justinbmeyer:
Maybe the guide should use the starter HTML similar to the main TodoMVC guide.  Start by wrapping it w/ the `<script>` tags so it becomes the template.  Then people are mostly just decorating it with `{{}}`.

chasenlehara:
I’ve updated the guide with instructions for how to start with the prototype HTML and modify it to create the app, à la the TodoMVC guide.

PR #2998, opened 2017-02-03
Adds can-observe as an ecosystem package

matthewp:
This adds can-observe as an ecosystem package. can-observe uses Proxys
to make plain objects observable.

See the docs to learn more:
https://github.com/canjs/can-observe/blob/master/docs/can-observe.md

PR #3148, opened 2017-04-04
Adds can-observe as an ecosystem package

matthewp:
This adds can-observe as an ecosystem package. can-observe uses Proxys
to make plain objects observable.

See the docs to learn more:
https://github.com/canjs/can-observe/blob/master/docs/can-observe.md


matthewp:
This replaces https://github.com/canjs/canjs/pull/2998

Issue #3117, opened 2017-03-20
Migration Guide: use String.prototype.trim instead of can.trim


mickmcgrath13:
`can.trim(str)` isn't available.  Should use `str.trim()` instead.

PR #3157, opened 2017-04-12
fixes #3115

### LINKED ISSUE #3115: Migration Guide: Correct typo for deep-assign
There is a typo in the docs: https://github.com/canjs/canjs/blame/master/docs/can-guides/commitment/migrating_to_3.0.md#L300

It should instead be import deepAssign from "can-util/js/`deep-assign/deep-assign`";

GeneralMeow:


matthewp:
👍

GeneralMeow:
should i get rid of my fix for #3117 ? doesn't seem to be passing the tests

docs/can-guides/components/todomvc-light.html
```diff
@@ -179,7 +179,7 @@
 			el.children('.edit').focus();
 		},
 		updateTodo: function (todo, el) {
-			var value = can.trim(el.val());
+			var value = String.prototype.trim(el.val());
```
chasenlehara: Should this be `String.prototype.trim.call(el.val())`?

chasenlehara: Also, I wonder if any protections for `null` or `undefined` need to be added… `String.prototype.trim.call` will throw if it’s given either of those values.

GeneralMeow:
Thank you so much for the feedback. I will incorporate it and get the PR updated!

docs/can-guides/components/todomvc-light.html
GeneralMeow: should i add .call on all the other String.prototype.trim functions as well?

PR #3158, opened 2017-04-12
Fix ValueConstructor behavior example

andrejewski:
Renamed `value => Value` and gave a potential example of a `Value` with a dummy `SubType`.


andrejewski:
This needs some with merge rights to merge, which I currently do not have.

PR #3169, opened 2017-04-14
Update can-event to 3.3.0

bmomberger-bitovi:
Also keep dependent packages in sync (fixed tests, fixed code in one case, and prevent multiple versions of can-event).


bmomberger-bitovi:
Working on it; this needs the latest can-stache-bindings to make tests pass.  If I update that, then can-ejs and can-legacy-view-helpers need to be updated, and right now I'm stuck on can-legacy-view-helpers because testee is breaking.

PR #3170, opened 2017-04-16
Adding can-stream tests

obaidott:
This PR adds support for:

- [x] can-stream
- [x] can-stream-kefir
- [x] can-define-stream
- [x] can-define-stream-kefir


phillipskevin:
@obaidott you should point this to minor.

package.json
```diff
@@ -49,8 +49,10 @@
     "can-construct-super": "3.0.3",
     "can-control": "3.0.7",
     "can-define": "1.0.17",
-    "can-define-stream": "0.0.7",
+    "can-define-stream": "github:canjs/can-define-stream#mixin",
```
phillipskevin: we should publish these first.

PR #3189, opened 2017-05-02
Add can-element as an ecosystem package

matthewp:


chasenlehara:
@matthewp I think the tests should also be added; not sure what else might be missing from the instructions on [Making a New Package](https://canjs.com/doc/guides/contributing/adding-ecosystem-modules.html).

PR #3241, opened 2017-05-11
Add image for each step of playlist-editor recipe

leoj3n:
See images folder for images to be added:

https://github.com/canjs/canjs/tree/playlist-editor-images/docs/can-guides/commitment/recipes/playlist-editor/images

I believe I have the correct syntax for the Markdown.

Hopefully this is helpful?


leoj3n:
it-works

PR #3254, opened 2017-05-15
Update the About page with images and more context

jamiemccue:
about-page


jamiemccue:
Removed CanJS from the title

Issue #250, opened 2013-01-23
Ignore attribute of Observe type while sending to server


wclr:
Solution to the problem proposed on forum:

<pre><code>

var Contact = can.Observe({
    attributes : {
        birthday : 'ignore'
    },
    serialize : {
        ignore : function(raw) {
            return undefined;
        }
    }
</code></pre>

This solution (return undefined) doesn't work if attribute is Observe (or Observe.List) as it seems that this serialize function is not called for Observes at all (as they have its own).

What can be other solution? I believe it is common situation when you don't want to send unnecessary fields to server. For example, you don't need it to be sent to server when object is saved when: 
1) you use working  property 'selected' or 'active' on Observe to identify that it is selected in the interface, . 
2) you have "readonly" fields that come from server and are not need to be send back.

Small "exessive" fields is ok, server "filter" will ignore it, but some times there are big nested objects that you actually ineffective to send over network if you really don't need them to be saved.

I could also override serialized method on Observe that that I don't want to be send, but it is universal and cumbersome solution, for me it would be better to have some static array of attributes that should be ignored while serialization/sending:

PR #3262, opened 2017-05-18
Add search! (and other minor upgrades)

mickmcgrath13:
* bump bit-docs-generate-html dependency version
  * https://github.com/bit-docs/bit-docs-generate-html/releases/tag/v0.5.0
  * https://github.com/bit-docs/bit-docs-generate-html/releases/tag/v0.6.0
* bump bit-docs-html-canjs dependency version
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.5.1
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.5.2
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.5.3
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.5.4
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.6.0
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.7.0
  * https://github.com/canjs/bit-docs-html-canjs/releases/tag/v0.7.1
* add bit-docs-generate-searchmap dependency
  * https://github.com/bit-docs/bit-docs-generate-searchmap/releases/tag/v0.0.1-0
* bump bit-docs-generate-searchmap dependency
  * https://github.com/bit-docs/bit-docs-generate-searchmap/releases/tag/v0.1.0
  * https://github.com/bit-docs/bit-docs-generate-searchmap/releases/tag/v0.1.1
* add steal-stache as a bit-docs html dependency


mickmcgrath13:
Some issues found so far:
* [ ] The `href` attributes in the search results aren't being generated properly
* [ ] When going from one page to another (pushState?), the search disappears

mickmcgrath13:
The href disappearance was due to improper scoping during the es6 -> es5 conversion (changed something to `self` that should have stayed `this`).

The search was disappearing because the `init` is run every time we "change pages" (pushState) - which re-initializes the search.  When the search is initialized, it makes an ajax request to `window.pathPrefix + '/searchMapHash.json'`.

..the problem is that the `window.pathPrefix` exists within a script tag on the newly ajax-retrieved page, but the value isn't making it onto the window (not being executed?) prior to the search being initialized.

While we should fix that (somehow?), it did highlight an opportunity which is: if we do not find the searchMapHash but we do have one in localStorage, we can just use that and continue as normal.

mickmcgrath13:
Maybe a fix could be instead of doing this:
```
<script type="text/javascript">
  window.pathPrefix = '{{pathToDest}}';
</script>
```

...we use a data attribute on a div:
```
<div data-path-prefix="{{pathToDest}}" />
```
..or on the body
```
<body data-path-prefix="{{pathToDest}}">
...
</body>
```

matthewp:
That solution sounds reasonable to me. I would check and see if the body gets updated when going from one page to another... if not putting it on a div might be a better way to go.

mickmcgrath13:
[Previous two issues](https://github.com/canjs/canjs/pull/3262#issuecomment-302528798) have solutions

Now, there's:
[ ] Responsive search

mickmcgrath13:
Issues mentioned above:
* [ ] [broken links](https://github.com/canjs/bit-docs-html-canjs/issues/249)
* [ ] [responsive search](https://github.com/canjs/bit-docs-html-canjs/issues/250)

mickmcgrath13:
https://github.com/canjs/bit-docs-html-canjs/pull/251

mickmcgrath13:
..so, now there's another issue... :/

With the latest update (#250 mentioned in the last comment), the `doc/` isn't at the beginning of the search results as it should be (seems to work in `bit-docs-html-canjs`...)

package.json
```diff
@@ -167,6 +175,6 @@
       "2.3.31": "https://v2.canjs.com"
     },
     "parent": "canjs",
-    "minifyBuild": true
+    "minifyBuild": false
```
chasenlehara: This will need to be flipped back to `true`

package.json
```diff
@@ -144,13 +144,21 @@
...
+      "templates": []
+    },
     "dependencies": {
       "bit-docs-glob-finder": "^0.0.5",
       "bit-docs-dev": "^0.0.3",
       "bit-docs-js": "^0.0.6",
       "bit-docs-tag-sourceref": "^0.0.3",
-      "bit-docs-generate-html": "^0.4.1",
-      "bit-docs-html-canjs": "^0.4.15",
+      "bit-docs-generate-html": "^0.5.0",
```
chasenlehara: `^0.6.0` now

justinbmeyer:
it seems like this checks in the `docMap.json`

justinbmeyer:
- the sidebar isn't dynamic width (it also is janky as the page is loading).
- the search doesn't appear to be in the right spot

cursor_and_canjs

- this can't be published without fixing the ability to show nested children:

cursor_and_canjs_-_api_docs

Are these structural changes in master?  If not, they should be reverted.  We shouldn't publish these structural changes accidentally.

chasenlehara:
The commits from this branch were merged but the search feature isn’t yet active; I will open a new PR to enable search sometime in the future.

Issue #2752, opened 2016-11-16
Homepage design


moschel:
Add design elements and content above the fold on the homepage.

Specifically, there are 3 types of content we might add, but we'll likely just pick two of them:
 * helloworld code sample in a jsbin
 * tiled buttons/images for chat guide links (Chat guide, TodoMVC guide, ATM guide)
 * images of apps using canjs and their company name

A designer should play with some options for including these, and then we'll decide what works best

tomgreever:
Here's one suggestion including the tiled buttons for the guides. 

Do we really want screenshots of the app? Or would company logos suffice? 

![screen shot 2016-11-16 at 11 06 02 am](https://cloud.githubusercontent.com/assets/326843/20357267/de71d3ce-abec-11e6-91c6-a8e2d7edca7d.png)

tomgreever:
@justinbmeyer or @brianmoschel can I get the hello world jsbin code example to include in this?

moschel:
I like the tiled guides links, looks like something I want to click. Tom, please use use any other jsbin for the purpose of this design exercise, we'll replace it if we move forward with the option

marshallswain:
> Do we really want screenshots of the app? Or would company logos suffice?

Company logos won't (likely) expire like apps do.

tomgreever:
Ok, I added a screenshot of a JSBin and moved things around. I'm uncertain if this is the look we're going for here, so any feedback from @justinbmeyer is welcome. I've not included company logos yet. Trying to avoid over designing it, so as a result everything is just kinda on the page. I'm not sure what to do next.
http://share.bitovi.com/TJOOEZ/#g=1&p=v3_home

![screen shot 2016-11-16 at 4 20 31 pm](https://cloud.githubusercontent.com/assets/326843/20368227/bf85db96-ac18-11e6-879b-f477cbf7f0e2.png)

andreaalmeida:
Updated homepage design:
- Re-organized side navigation pages (mockup included)
- Added border around turtle hare images so that they look more like one image, instead of two separate ones
- Moved NPM Package, Github Star, and GitHub edit links/buttons in-line with page title
- Styled Edit page in Github link to look like a button
- Removed collections and packages sections (now live on API docs page)
- Added logos of companies that use CanJS
- Increased font-weight of "on this page" to make it more noticeable 


[...]

PR #3284, opened 2017-05-26
Promoted Recipes page

### LINKED ISSUE #3227: Promote the Recipes page to one level higher
Instead of being under its current _commitment_ section, it should be its own section.

imaustink:
image


chasenlehara:
This is close but not quite right. The sitemap in [this comment](https://github.com/canjs/canjs/issues/2752#issuecomment-299529110) makes it a little bit more clear that the goal is to show all the recipes in the nav without having to click on Recipes. It should come out to something like this:
screen shot 2017-05-26 at 2 10 43 pm

imaustink:
Updated the PR :)

docs/can-guides/Guides.md
```diff
@@ -1,10 +1,11 @@
 @page guides Guides
 @parent canjs 1
 @group guides/getting-started 0 getting started
-@group guides/experiment 1 experiment
-@group guides/commitment 2 commitment
-@group guides/contribute 3 contribute
-@group guides/upgrade 4 upgrade
+@group guides/experiment 2 experiment
+@group guides/commitment 3 commitment
```
chasenlehara: I think this should be 4. Right now I’m seeing some weird behavior when switching between pages:
recipes nav bug

imaustink:
Fixed 😃 
canjs-side-nav

Issue #3285, opened 2017-05-26
can.batch callback not executed if registered during batch dispatching


UselessPickles:
Problem was discovered in v2.3.27. Appears to still exist in v2.3.31.

If a can.batch callback is registered via can.batch.after() while a batch is currently dispatching, the callback is added to the batch's list of callbacks, but is never called after the batch dispatching is complete.

This is most clearly explained with a simple demo of the problem:

```
    var DataModel = can.Map.extend({
        define: {
            value: {
                type: "compute"
            }
        }
    });

    var dataModel = new DataModel({
        value: 42
    });

    dataModel.on("value", function() {
        can.batch.after(function() {
            alert("I'm never called!");
        });
    });

    dataModel.attr("value", 2);

```

The cause of the problem is in batch.js, in the "stop" method. While processing all the batches, the batch's callbacks are pushed onto a local callbacks array (to gather up ALL callbacks). This is done BEFORE dispatching the batch's events, though. So any callbacks that are registered from within the event handlers are never added to the local list of all callbacks.

```
    var callbacks = [], i;
    while (batch = batches.shift()) {
        var events = batch.events;
        callbacks.push.apply(callbacks, batch.callbacks); // <--- PROBLEM!
        dispatchingBatch = batch;
        can.batch.batchNum = batch.number;
        var len;
        if (callStart) {
            can.batch.start();
        }
        for (i = 0, len = events.length; i < len; i++) {
            can.dispatch.apply(events[i][0], events[i][1]);
        }
        can.batch._onDispatchedEvents(batch.number);
        // callbacks.push.apply(callbacks, batch.callbacks) <--- SOLUTION!
        dispatchingBatch = null;
        can.batch.batchNum = undefined;
    }
    for (i = callbacks.length - 1; i >= 0; i--) {
        callbacks[i]();
    }
```

phillipskevin:
@UselessPickles I know I commented on your PR already, but I want to make sure I fully understand the issue, is there a reason you're using `can.batch.after`?

UselessPickles:
[...]

PR #3286, opened 2017-05-26
Fix can.batch.stop() to call ALL callbacks

UselessPickles:
Fixes this issue:  https://github.com/canjs/canjs/issues/3285

### DIFF
--- a/test/all-tests.js
+++ b/test/all-tests.js
@@ -34,4 +34,4 @@ steal("can/test/test.js",
 		"can/view/scope/scope_test.js",
 		"can/util/string/string_test.js",
 		"can/util/attr/attr_test.js",
-    "can/util/batch/batch_test.js", function() {});
+		"can/util/batch/batch_test.js", function() {});
--- a/test/all-tests.js
+++ b/test/all-tests.js
@@ -33,4 +33,5 @@ steal("can/test/test.js",
 		"can/view/live/live_test.js",
 		"can/view/scope/scope_test.js",
 		"can/util/string/string_test.js",
-		"can/util/attr/attr_test.js", function() {});
+		"can/util/attr/attr_test.js",
+    "can/util/batch/batch_test.js", function() {});
--- a/util/batch/batch.js
+++ b/util/batch/batch.js
@@ -202,7 +202,7 @@ steal('can/util/can.js', function (can) {
 
 						can.batch._onDispatchedEvents(batch.number);
 						
-            // NOTE: callbacks must be gathered up AFTER dispatching all events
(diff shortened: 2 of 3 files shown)


UselessPickles:
I'm sure it's obvious, but I'm new to github and contributing to this project. I clearly am not doing something right because my simple "move a line and add a comment" change is failing automated tests. Looks like it's probably a discrepancy in what character(s) is used for the newline character. 

I made the change directly on github with the web-based file editor. Thought I could get away with a simple change like this without setting up a local environment, etc. 

If someone else could corrected and moved through, or give me some clear guidance on what I need to do, I'd really appreciate it. This bug is currently a critical roadblock to a project that is already running behind schedule for me.

phillipskevin:
I think you might be able to fix it by changing the wrapping in the editor:
image

However, it might be a good idea to set up the repo locally anyway. It would be great if you could add a test for this case. It would basically be the same as the example you gave in the issue, but with an assertion (`ok(true, ...)`) in place of the `alert`.

UselessPickles:
I tried running unit tests locally via "grunt test", but got this error:

Running "testee:steal" (testee) task
  Fatal error: spawn D:\DevProjects\github\canjs\node_modules\.bin\phantomjs ENOENT

Any ideas what I did wrong? This is on a windows 10 machine. I installed git, NodeJS, grunt-cli, cloned by canjs repo, checked out my "patch-1" branch, ran "npm install" and "bower install" in my repo directory. Am i missing a step?

For now I pushed my changes to see if my new unit test passes.

UselessPickles:
Also, looks like I still have some whitespace issues, probably caused by an inconsistency between using space characters or tab characters for indentation. What's the standard for this project?

UselessPickles:
Now that the tests have passed, is there any way to confirm that my new unit test actually executed? I can't find anything in the logs.

phillipskevin:
We're using tabs. You can set up EditorConfig (http://editorconfig.org/) in your editor to make sure your spacing is correct.

Instead of using the grunt task to run the tests, you can just load one of the test pages in a browser (`test/jquery.html` for instance).

UselessPickles:
@phillipskevin All indents converted from spaces to tabs. As far as I can tell, my new unit test should be executing and is passing, but I still am not sure where to look for some sort of log that confirms that the new test in batch_test.js file specifically was executed/passed.

UselessPickles:
Is this ever going to be merged? If not, I'd like to close it to clean up my list of pending PRs.

justinbmeyer:
I'll try to find someone to take a look at it tomorrow.  Thanks for being persistent!

phillipskevin:
This is published in `can@2.3.34`.

PR #3301, opened 2017-06-07
fix(package): update can-util to version 3.8.4

phillipskevin:
Closes #3280


phillipskevin:
There have been a few other repos with failing tests with can-util updates, so running this as a sanity check.

PR #3304, opened 2017-06-08
Add search

chasenlehara:
This PR is ready to be reviewed. There are some [remaining issues](https://github.com/canjs/bit-docs-html-canjs/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20search) but I don’t think any of them are blockers.


justinbmeyer:
btw @chasenlehara  / @imaustink this site shouldn't have to be hosted from `/` to work.  I host from a parent folder of `canjs` and it isn't working.  I noticed the images in about are failing:


cursor_and_canjs_-_about

imaustink:
@justinbmeyer I've updated the PR. It should allow hosting it 'relatively' and fix the images on the about page. I also have [PR](https://github.com/canjs/bit-docs-html-canjs/pull/325) in to fix the signature parameters.

chasenlehara:
CI is failing because of conflicting `can-types` versions. [can-connect](https://github.com/canjs/can-connect/blob/1cdd9f883217274e25213228125e4e08518bc5d0/package.json#L25) and [can-view-import](https://github.com/canjs/can-view-import/blob/6b746a5f4a071a305899ae8b4d53c42daa3d6d64/package.json#L37) both require `steal-stache`, which has been updated for `can-reflect`, so its dependencies require `can-types@1.1`, which conflicts with the `can-types` version of the other dependencies in this repo.

So, a couple questions:

1) Is it necessary for `can-connect` and `can-view-import` to depend on `steal-stache`? Could it instead be a dev-dependency?
2) Is it ok to merge this PR as-is with the build failing because the changes in the PR aren’t causing the failure?

I believe this is the same issue @bgando ran into with her PR https://github.com/canjs/canjs/pull/3370

PR #3306, opened 2017-06-08
updating to steal 1.0

phillipskevin:


christopherjbaker:
When running locally, I get an error of `Uncaught TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.` (the exact test that fails varies) in both dev and production (had to add `steal-css` and change the steal.prod.js path in production.html to get it to build and run in production). I narrowed that error down to being called on a `<p>` created in a test, and if I disable [the test that creates it](https://github.com/canjs/can-stache-bindings/blob/v3.1.5/test/bindings-test.js#L2969-L2989) by switching it to `QUnit.skip`, everything passes fine. I cannot see why this would cause that error, and why it would happen so much later. Also, if that error is caught and ignored, all tests pass too.

phillipskevin:
Good catch on the steal-css and steal.prod.js.

That test is failing because those tests are switching between the normal dom and vdom. So it's dispatching that event, then the dom is switching, then the event handler is being called. I think a simple thing we should try is adding a `stop()`/`start()` to that test.

christopherjbaker:
Adding start/stop to that test did not make a difference, but a [new empty test with a short delay](https://github.com/canjs/can-stache-bindings/pull/227) did. With that change, this passes all tests locally.

bmomberger-bitovi:
I've wrestled with this same error at times in the past, and also wrote some code mitigating it happening more often (by making another mutation fire and waiting for it), and I think a lot of it comes down to mutation events not having a fixed time or ordering when they fire.  Browser engines are free to batch these events and dispatch them as they see fit.  There's no guarantee that a mutation event will happen within a time box.  For the most part, outliers don't happen, but it seems like we cross some threshold at some point where the browser has to do an allocation or a buffer reset or something, and then we get one mutation batched up way out of order.

For now, use whatever tricks you can to make the test pass.  The code itself is tight; it's only an issue because the test battery has to try DOM and vdom in succession.  If we can fix things with a delay, do it.  But I'd also extensively document what you're doing there, so if and when we have the issue again, we won't have to rely on our short memories to inform us of what's going on.

christopherjbaker:
fixed in #3381

PR #3315, opened 2017-06-16
Rename API guide page. Closes #2863

### LINKED ISSUE #2863: Move the API Guide page
The [API Guide](http://canjs.com/doc/guides/api.html) currently has this outline:

1. Documentation Structure
2. Package and Module Pages
3. Functions, Properties, and Typedef pages
4. How to find what you’re looking for…

When UX testers went to this page, they expected to see an overview of the entire CanJS API, not info about how to use the site.

Arguably, we shouldn’t need to have a page that tells the users about the site’s information architecture; it should be obvious how to move through the site without reading a how-to guide. However, if we do keep this page, we should either rename it or move it so new users aren’t confused by its presence.

imaustink:
image

### DIFF
--- a/docs/can-guides/Guides.md
+++ b/docs/can-guides/Guides.md
@@ -1,10 +1,10 @@
 @page guides Guides
 @parent canjs 1
-@group guides/getting-started 0 getting started
-@group guides/experiment 2 experiment
-@group guides/commitment 4 commitment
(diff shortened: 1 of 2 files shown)


chasenlehara:
Closing in favor of https://github.com/canjs/canjs/pull/3450

PR #3319, opened 2017-06-19
Add ReactViewModel

christopherjbaker:
I added the shrinkwrapped dependencies, modified the docs glob, and incorporated the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ publish-docs:
 	git add -f node_modules/jquery-ui
 	git add -f node_modules/funcunit
 	git add -f node_modules/syn
+	git add -f node_modules/react
+	git add -f node_modules/react-dom
+	git add -f node_modules/react-view-model
 	git add -f node_modules/socket.io-client
 	git add -f node_modules/feathers/package.json
 	git add -f node_modules/feathers-authentication-client/package.json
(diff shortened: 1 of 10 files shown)


christopherjbaker:
Current test failures are fixed in canjs/can-stache-converters#34

justinbmeyer:
@m-mujica / @matthewp it seems that `process` isn't provided anymore to the built files.  Can you take a look at this?

matthewp:
Are you talking about what is loaded by test/index.html? That loads can.all.js, which is a global exported build. As far as I remember the global shim never had `process`, only the steal npm plugin (which does still have it). I'm not opposed to adding it to the global shim though.

matthewp:
I don't understand how this is supposed to work.  There are several libs being ignored in the build: https://github.com/canjs/canjs/blob/master/build.js#L6-L7 . but there are no corresponding script tags for those libs in the text HTML pages. So I don't understand how those libs are not throwing.

matthewp:
Digging into it more it looks like, in the example of can-stream-kefir, that kefir is excluded from the build and is not on the window, but because can-stream-kefir doesn't care in the module body if `Kefir` is undefined, it doesn't break. But it would break if someone attempted to use it (although I'm not quite sure how they would do so).

matthewp:
So I think we need to either include react and all of its dependencies into the build or exclude it, and react-view-model and can-react-component. For now I'm not to exclude those.

justinbmeyer:
Yeah, we should do it like kefir and jQuery ... not include them in the build, while allowing the build to work without them (at least the modules that don't depend on them).

I know for jQuery, we do this conditionally a few places. We test that the jQuery module actually provided a value.

PR #3321, opened 2017-06-21
Fix qunit ok

### LINKED ISSUE #3276: Fix incorrect uses of QUnit.ok
When QUnit.ok is called with incorrect arguments like `QUnit.ok(m.l[0].id, 5, "correct props");` an error is thrown:

image

One example can be found [here](https://github.com/canjs/can-define/blob/master/test/test.js#L70), but this should be changed in all can-* projects. To track down all occurrences, we could overwrite `QUnit.ok` to throw an error when `arguments > 2`.

We should correct these to correctly use `QUnit.equal`.

See https://github.com/bitovi/testee/issues/136 for more details.

obaidott:
Fixes #3276


andrejewski:
Superseded by https://github.com/canjs/canjs/pull/3434

PR #3326, opened 2017-06-22
Update credit-card-simple.md

leoj3n:
Some tweaks that I think help make the guide easier to parse through.


leoj3n:
Another part I feel is a bit confusing for a beginner is the way this mentions "if/else branching" but doesn't really show that in it's entirety:

https://github.com/canjs/canjs/pull/3326/files#diff-e82e0aad7bb0050545316a9e3d28f4dfR280

PR #3365, opened 2017-06-24
Add can-ndjson-stream to ecosystem docs

bgando:


package.json
```diff
@@ -64,6 +65,7 @@
     "can-map-backup": "3.0.3",
     "can-map-define": "3.0.9",
     "can-namespace": "1.0.0",
+    "can-ndjson-stream": "file:../can-ndjson-stream",
```
nlundquist: should probably change this to a published version number before merging

package.json
```diff
@@ -39,12 +39,13 @@
     "github": "https://github.com/canjs/canjs"
   },
   "dependencies": {
+    "bit-docs-generate-html": "^0.6.0",
```
nlundquist: not sure this is needed since it's part of the bit-docs dependencies

package.json
```diff
@@ -95,7 +97,7 @@
     "can-zone": "0.6.7"
   },
   "devDependencies": {
-    "bit-docs": "^0.0.7",
+    "bit-docs": "0.0.7",
```
chasenlehara: Why was this changed?

chasenlehara:
Closing this in favor of https://github.com/canjs/canjs/pull/3370

bgando:
This is a standalone module that #3370 (can-connect-ndjson) depends on but is not the same

PR #3370, opened 2017-06-29
Add can-connect-ndjson and can-ndjson-stream docs and tests

bgando:


package.json
```diff
@@ -39,12 +39,14 @@
     "github": "https://github.com/canjs/canjs"
   },
   "dependencies": {
+    "bit-docs-generate-html": "^0.6.0",
```
nlundquist: this is already listed in the bit-docs dependencies so this wouldn't need to be here

package.json
```diff
@@ -95,7 +98,7 @@
     "can-zone": "0.6.7"
   },
   "devDependencies": {
-    "bit-docs": "^0.0.7",
+    "bit-docs": "0.0.7",
```
chasenlehara: Why was this changed?

bgando:
I am still experiencing these errors, but I do not believe it is related to can-connect-ndjson or can-ndjson-stream.
image

Issue #3369, opened 2017-06-27
Remove setValue from Maps


justinbmeyer:
Related branches:

- [ ] - [can-reflect](https://github.com/canjs/can-reflect/pull/26) - `1.1.0-pre.3`
- [ ] - [can-observation](https://github.com/canjs/can-observation/pull/94) - `3.3.0-pre.2`
- [ ] - [can-compute](https://github.com/canjs/can-compute/pull/86) - `3.3.0-pre.1`
- [ ] - [can-reflect-promise](https://github.com/canjs/can-reflect-promise/pull/6) - `1.1.0-pre.0`
- [ ] - [can-stache-key](https://github.com/canjs/can-stache-key) - `0.0.2`
- [ ]  - [can-define](https://github.com/canjs/can-define/pull/226) - `1.3.0-pre.2`
- [ ] - [can-map](https://github.com/canjs/can-map/pull/49) - `3.3.0-pre.1`
- [ ] - [can-list](https://github.com/canjs/can-list/pull/43) - `3.2.0-pre.2`
- [ ] - [can-simple-map](https://github.com/canjs/can-simple-map/pull/36) - `3.3.0-pre.2`
- [ ] - [can-view-scope](https://github.com/canjs/can-view-scope/pull/67) -`3.3.0-pre.3`
- [ ] - [can-jquery](https://github.com/canjs/can-jquery/pull/91) - `3.2.0-pre.0`
- [ ] - [can-view-model](https://github.com/canjs/can-view-model/pull/39) - `3.4.0-pre.0`
- [ ] - [can-route](https://github.com/canjs/can-route/pull/86) - `3.2.0-pre.0`
- [ ] - [can-view-callbacks](https://github.com/canjs/can-view-callbacks/pull/45) - `3.2.0-pre.0`
- [ ] - [can-view-live](https://github.com/canjs/can-view-live/pull/40) - `3.2.0-pre.0`
- [ ] - [can-stache](https://github.com/canjs/can-stache/pull/228) - `3.2.0-pre.0`
- [ ] - [can-stache-bindings](https://github.com/canjs/can-stache-bindings/pull/245) - `3.4.0-pre.0`
- [ ] - [can-control](https://github.com/canjs/can-control/pull/72) `3.2.0-pre.0`
- [ ] - [can-component](https://github.com/canjs/can-component/pull/113) `3.2.0-pre.0`

make map assign faster ... (ignore current)

PR #3381, opened 2017-06-30
CanJS 3.9

justinbmeyer:
Changes listed in #3369


justinbmeyer:
@christopherjbaker looks like this still needs the integration test enabled.  I'm going to flip that switch in a moment.

PR #3393, opened 2017-07-05
MERGE ME: 3.9

justinbmeyer:


docs/can-guides/contribute/including-new-modules.md
```diff
@@ -1,21 +1,81 @@
-@page guides/contributing/adding-ecosystem-modules New Package
+@page guides/contributing/new-package New Package
```
chasenlehara: I’d leave the slug alone so we don’t break old links to this page.

justinbmeyer: the old slug creates ordering problems I'll check for broken links before I release

PR #3505, opened 2017-08-24
Really add can-observe to ecosystem

imaustink:
Closes canjs/can-observe#7


matthewp:
Are the doc links and github stars stuff added?

imaustink:
@matthewp they are now

matthewp:
@imaustink I think you need to wrap the require() in the test in an if statement, checking for `typeof Proxy === "function"`

imaustink:
@phillipskevin should I base this off `minor` instead of `master` then?

phillipskevin:
why do you need to update the other dependencies? it would be nice to get this in now instead of waiting until we're ready for 3.12.0.

imaustink:
The build won't pass without upgrading those deps. Mostly upgrading just removed `can-util/log` and `can-util/dev` but I think there was a least 1 other fix as well.

phillipskevin:
Ok, then can you point this to the `minor` branch instead? We can try to get a `3.12.0` out tomorrow or early next week.

PR #3506, opened 2017-08-25
Added instruction to Setting Up CanJS section.

MikoBrandini:
A top level property must be added manually to the package.json file in order for the instructions to work. It is not added automatically when you install the packages.


chasenlehara:
Awesome, thank you for catching this and opening a PR @MikoBrandini!

PR #3513, opened 2017-08-30
Register steal-css as a steal-plugin in package.json

### LINKED ISSUE #3490: can-view-callbacks demo doesn’t work
From https://canjs.com/doc/can-view-callbacks.attr.html

screen shot 2017-08-14 at 3 00 22 pm

Filing in this repo because this is where the demo files are: https://github.com/canjs/canjs/tree/master/demos/can-view-callbacks

bmomberger-bitovi:
This fixes #3490 by making steal-css available as a plugin in demo frames.  Without it, .css file paths aren't identifies as CSS resources, and are instead treated as modules with implicit .js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,7 +158,10 @@
       "socket.io-client/dist/socket.io": {
         "format": "cjs"
       }
-    }
+    },
+    "plugins": [
+      "steal-css"
+    ]
   },
   "bit-docs": {
     "html": {


phillipskevin:
This is a problem because `steal-css` is a devDependency? So is the correct solution just to use a `<link>` for these css files: https://github.com/canjs/canjs/blob/a181a6a6b294e9e8d29a5aaa3f761f6cb3c4481f/demos/can-view-callbacks/fade_in_when.html#L20.

bmomberger-bitovi:
Closing in favor of a #3519

PR #3558, opened 2017-09-13
adding bower install to legacy publishing docs

phillipskevin:


contributing.md
```diff
@@ -125,7 +125,7 @@ After you have completed those steps simply run `grunt build` and it will put th
 ## Publishing & Releasing
 
 1. Install and use Node.js 0.12 ([nvm](https://github.com/creationix/nvm) is useful for managing multiple versions locally)
-2. Install the dependencies: `npm install && npm install grunt-cli -g`
+2. Install the dependencies: `npm install && npm install bower -g && bower install && npm install grunt-cli -g`
```
chasenlehara: I wonder if we could get away with installing bower globally if the command was something like `./node_modules/.bin/bower install`

phillipskevin: yeah, we could do that with `grunt-cli` also if we care enough. I don't really care what I have installed globally on node 0.12 since I only ever use it for releasing legacy can 2.3. What do you think?

chasenlehara: Same, I don’t really care, the only advantage I see is that both of those would be tied to what’s in the `package.json` instead of the latest version of those packages.

phillipskevin: I'm just going to leave it this way for now. There are several grunt commands and I don't think it's worth taking the time to update it.

PR #3573, opened 2017-09-22
Update theme to include survey ad and IE fixes

chasenlehara:
Closes https://github.com/canjs/canjs/issues/3548


chasenlehara:
@imaustink Thanks! That’s what I get for changing the text last minute…

PR #3581, opened 2017-09-25
add can-ajax docs

cherifGsoul:
this solves https://github.com/canjs/can-ajax/issues/9


cherifGsoul:
@phillipskevin the requested changes are commited thank you for checking :)

cherifGsoul:
@phillipskevin the tests fail because of a time out error, all tests pass localy !!

cherifGsoul:
I think the test folder in ```can-ajax``` npm module is needed the tests needs them

test/test-production.js
```diff
@@ -29,6 +29,7 @@ require('can-symbol/can-symbol-test');
 require('can-reflect/can-reflect-test');
 require('can-reflect-promise/test/can-reflect-promise_test');
 require('can-attribute-encoder');
+require('can-ajax');
```
phillipskevin: This should be the test file. Also.. the one above this is wrong, should import the tests for `can-attribtue-encoder` if you want to fix that up also.

package.json
```diff
@@ -65,6 +66,7 @@
     "can-jquery": "3.2.1",
     "can-kefir": "0.2.1",
     "can-list": "3.2.1",
+    "can-make-map": "^1.0.0",
```
phillipskevin: This should be a devDependency.

cherifGsoul: yes

PR #3615, opened 2017-10-04
Update cta-bus-map.md

leoj3n:
- Hide JS editor of final example.
- Consistently capitalize list items.
- Change the "what's provided" tone for better clarity.
  - Before it sounded like steps that should be followed.

### DIFF
--- a/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
+++ b/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
@@ -12,9 +12,11 @@ In this guide you will learn how to:
 - Create a custom element that wraps a google map.
 - Add markers to the google map.
 
+## Final Widget Example ##
+
 The final widget looks like:
 
-<a class="jsbin-embed" href="https://jsbin.com/zewenov/4/embed?js,output&height=600px">JS Bin on jsbin.com</a>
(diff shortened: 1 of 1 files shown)


docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
```diff
@@ -68,7 +62,7 @@ __A Basic CanJS Setup__
   ```
 
 - CanJS uses [can-stache] to render data in a template
-  and keep it live.  Templates can be authored in `<script>` tags like:
+  and keep it live.  Templates are authored in `<script>` tags like:
```
justinbmeyer: they shouldn't be in a real app.

justinbmeyer:
Thanks for suggestions.  Incorporated into another commit.

leoj3n:
image

Think this looks better / less confusing.

docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
leoj3n: well, they ARE authored in `<script>` tags _for this example_ which is why I found this sentence confusing — perhaps you could add a short blurb to qualify that a real app would organize templates into files.

PR #3723, opened 2017-11-07
Fix selected-to-index demo

imaustink:
Fixed canjs/can-stache-converters#61


imaustink:
@phillipskevin You're correct, I missed that and thought it was a typo. I've updated the PR to add a new demo for `selected-to-index`.

Issue #3793, opened 2017-12-07
Check the demos for 4.0


chasenlehara:
- Update any in 3.0 that are using old/deprecated syntax
- Remove any that are no longer used
- Make sure demos are working in 4.0 site

matthewp:
Which demos? @James0x57 and I did the recipes and TodoMVC and that sort of thing.

chasenlehara:
There are also demos in all the individual packages that need to be checked.

justinbmeyer:
I don't know if many packages have demos that are actually shown in the docs.  The `demos` folder is the most important one here.

chasenlehara:
This was completed with these PRs:

- https://github.com/canjs/canjs/pull/3799
- https://github.com/canjs/canjs/pull/3800
- https://github.com/canjs/canjs/pull/3803
- https://github.com/canjs/canjs/pull/3804
- https://github.com/canjs/canjs/pull/3805
- https://github.com/canjs/canjs/pull/3806
- https://github.com/canjs/canjs/pull/3807
- https://github.com/canjs/canjs/pull/3811
- https://github.com/canjs/canjs/pull/3812

PR #3796, opened 2017-12-08
Update the can-component demos with new CanJS 3 features

chasenlehara:
This avoids using deprecated APIs that will be removed in CanJS 4 and fixes some other warnings.

It also removes some files that weren’t being used.

Part of https://github.com/canjs/canjs/issues/3793 [there will be more demo updates]


chasenlehara:
Based on our conversation during the contributors meeting, I’m going to update this to explicitly use `scope.root` to avoid `../` in our demos.

demos/can-component/selected.html
```diff
@@ -11,8 +11,8 @@
 <div id='out'></div>
 <script type="text/stache" id="demo-html">
 {{#items}}
-  <div {{#isSelected}}class='selected'{{/isSelected}}
-       on:click='toggle'>
+  <div {{#isSelected()}}class='selected'{{/isSelected}}
+       on:click='toggle(.)'>
```
phillipskevin: I would use `toggle(this)`.

demos/can-component/treecombo.html
```diff
@@ -49,11 +49,11 @@
   <ul class='options'>
     <content>
       {{#selectableItems}}
-        <li {{#isSelected}}class='active'{{/isSelected}} on:click='toggle(.)'>
-          <input type='checkbox' {{#isSelected(.)}}checked{{/isSelected}}/>
+        <li {{#scope.root.isSelected(.)}}class='active'{{/isSelected}} on:click='toggle(.)'>
```
phillipskevin: Same thing here... let's show `this` instead of `.`.

PR #3799, opened 2017-12-12
More demo updates

chasenlehara:
- Add `demos/can-stache-bindings-legacy/`
- Update the can-stache-bindings demos with new CanJS 3 features
- Fix the can-stache-converters `select[multiple]` demo
- Update can-view-callbacks demos with new syntax

### DIFF
--- a/demos/can-stache-converters/multi-values.html
+++ b/demos/can-stache-converters/multi-values.html
@@ -1,37 +1,31 @@
-<!doctype html>
-<html>
-<head>
-	<title>values!</title>
-	<style>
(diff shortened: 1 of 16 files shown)


demos/can-stache-converters/multi-values.html
```diff
@@ -44,24 +38,21 @@
 </select>
 
 <ul>
-	{{#each(colors)}}
+	{{#each colors}}
```
phillipskevin: why is this one using helper syntax?

Issue #3760, opened 2017-11-30
Add can-route & can-route-pushstate tests [3.0 & 4.0]


chasenlehara:
In 3.0, the `can-route` & `can-route-pushstate` tests are not being run: https://github.com/canjs/canjs/blob/9e73bda37ed4ecdacc0375586b4a10cdf375d06b/test/test.js#L7-L8

I discovered this while working on the 4.0 build. They were probably commented out because they were broken if added to this repo.

This change in `can-route` might need to be backported from 4.0 to 3.0 to make the tests work: https://github.com/canjs/can-route/commit/0be7e2188259d7ef2e56c6e15d66ce0c9ccfcaf6

This test in `can-route-pushstate` loads an iframe at a url that doesn’t match the path in this repo, so it doesn’t load correctly: https://github.com/canjs/can-route-pushstate/blob/736328b8b496b11070a837cde3e806eda1ab039f/can-route-pushstate_test.js#L450

matthewp:
Can this be closed?

chasenlehara:
It looks like the tests were added to the 4.x suite but not for 3; not sure how much we care about making sure the tests are running there too.

PR #3801, opened 2017-12-13
add the route and pushstate tests

James0x57:
For #3760 

image


chasenlehara:
I think `route/route_test.js` should point to `can-route/test/test`

James0x57:
Windows 10 Chrome is failing in Travis.

They work on browserstack win 10 chrome:
image

And they all pass on my windows 10 laptop in chrome latest too.

PR #3807, opened 2017-12-15
More fixes for the demos

chasenlehara:


demos/can-util/input-focused.html
```diff
@@ -23,23 +23,24 @@
...
-			putInFocus: function(){
-				this.inFocus = true;
-			}
-		});
-		var map = new ViewModel({ inFocus: false });
-		window.map = map;
+<script src="../../node_modules/steal/steal.js" main="@empty">
+var stache = require("can-stache");
+var DefineMap = require("can-define/map/map");
+require("can-stache-bindings");
```
chasenlehara: Switching from requiring `can-stache-converters` to just `can-stache-bindings` was the only change here.

demos/can-component/selected.html
```diff
@@ -12,7 +12,7 @@
 <script type="text/stache" id="demo-html">
 {{#items}}
   <div {{#isSelected()}}class='selected'{{/isSelected}}
```
phillipskevin: I don't think this will work (in 4.0)... `isSelected()` won't be passed an `options` argument. You should just pass the item in like `isSelected(this)`

demos/can-view-callbacks/dynamic_tooltip.html
```diff
@@ -24,13 +24,13 @@
 		<tr>
 			<td colspan="3">
 				<button tooltip="{{deleteTooltip()}}"
-						{{^selected.length}}class="disabled"{{/selected.length}}
+						{{^if(selected.length)}}class="disabled"{{/if}}
```
phillipskevin: did `{{^selected.length}}` not work?

demos/can-util/input-focused.html
phillipskevin: FYI `can-stache-converters` requires `can-stache-bindings`, but this is fine since no converters are being used.

chasenlehara: Yeah, which is why it was working, but I couldn’t see a reason why converters needed to be required…

demos/can-view-callbacks/dynamic_tooltip.html
chasenlehara: I think it works but generates warnings if `selected` is `undefined`, so I thought this would be a better way of writing the same thing.

phillipskevin: Ok, we've got an issue open to fix that. Thanks.

PR #3811, opened 2017-12-19
More demo updates

chasenlehara:
The last of my updates to the 3.0 demos. Maybe.

### DIFF
--- a/demos/can-component/paginate.html
+++ b/demos/can-component/paginate.html
@@ -264,7 +264,7 @@ var GridViewModel = DefineMap.extend({
 				resolve(false);
 			});
 
-			resolve(true);
+			return true;
 		}
 	}
 });
--- a/demos/can-component/paginate.html
+++ b/demos/can-component/paginate.html
@@ -74,7 +74,7 @@ var DefineList = require("can-define/list/list");
 var Component = require("can-component");
 var fixture = require("can-fixture");
 var stache = require("can-stache");
-var superMap = require("can-connect/can/super-map/super-map");
+var baseMap = require("can-connect/can/base-map/base-map");
 var set = require("can-set");
 var compute = require("can-compute");
 
@@ -97,7 +97,7 @@ var websiteAlgebra = new set.Algebra(
 	set.props.offsetLimit("offset", "limit")
 );
 
-superMap({
+baseMap({
 	idProp: "id",
 	Map: Website,
 	List: Website.List,
@@ -246,8 +246,7 @@ var AppViewModel = DefineMap.extend({
 
 Component.extend({
 	tag: "my-app",
-	ViewModel: AppViewModel,
-	leakScope: true
+	ViewModel: AppViewModel
 });
 
 var GridViewModel = DefineMap.extend({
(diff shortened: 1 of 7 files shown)


demos/can-component/paginate.html
```diff
@@ -97,7 +97,7 @@
 	set.props.offsetLimit("offset", "limit")
 );
 
-superMap({
+baseMap({
```
chasenlehara: Switching to base-map because the can-fixture delay wasn’t perceivable; I think because the data was being fetched from local-storage before the fixture would resolve.

demos/can-component/paginate.html
```diff
@@ -246,8 +246,7 @@
 
 Component.extend({
 	tag: "my-app",
-	ViewModel: AppViewModel,
-	leakScope: true
```
chasenlehara: I don’t think `leakScope: true` was necessary here.

demos/can-component/paginate.html
```diff
@@ -264,14 +263,15 @@
 			promise.then(function() {
 				resolve(false);
 			});
-			return true;
+
+			resolve(true);
```
chasenlehara: As far as I can tell, the docs say that return values are ignored if a setter has two arguments, so I don’t think this ever worked.

demos/can-component/paginate.html
```diff
@@ -264,14 +263,15 @@
...
 		}
 	}
 });
 
 Component.extend({
 	tag: "my-grid",
-	view: stache("<table><tbody {{#waiting}}class='waiting'{{/waiting}}><content></content></tbody></table>"),
+	view: stache("<table><tbody {{#if(waiting)}}class='waiting'{{/if}}><content></content></tbody></table>"),
```
chasenlehara: Avoiding can-stache’s “unable to find key” warning

demos/can-component/tabs.html
```diff
@@ -49,7 +49,10 @@
 var canBatch = require("can-event/batch/batch");
 
 var Panel = DefineMap.extend({
-	active: "boolean"
+	active: {
+		type: "boolean",
+		value: false
+	}
```
chasenlehara: Also avoiding can-stache’s “unable to find key” warning

demos/can-stache-bindings/event-args.html
```diff
@@ -1,6 +1,6 @@
 <script type='text/stache' can-autorender id='demo-html'>
-{{#each(items)}}
-	<li on:click='items.splice(scope.index,1)'>{{name}}</li>
+{{#each(items, item=value num=index)}}
+	<li on:click='../items.splice(num,1)'>{{item.name}}</li>
```
chasenlehara: I don’t know if it’s a bug or not, but `scope.index` can’t be used because it’ll be the last value after going through the entire list.

phillipskevin: You mean it's always set to `items.length - 1`?

phillipskevin: If so, that's a bug.

demos/can-component/paginate.html
phillipskevin: I think this is fine, but it should work the other way also: https://jsbin.com/gikewuyewi/1/edit?html,js,output.

PR #3812, opened 2017-12-20
4.0 demo updates

chasenlehara:
- Remove unnecessary use of @ in stache template

### DIFF
--- a/demos/can-component/accordion.html
+++ b/demos/can-component/accordion.html
@@ -1,14 +1,14 @@
 <div id='out'></div>
 <script id="demo-html" type="text/stache">
 <my-accordion>
-	<my-panel title:from="'Contacts'">
+	<my-panel makeActive:from="makeActive" title:from="'Contacts'">
 		<button>Change title attribute to "Users"</button>
 		<ul>
 			<li>Justin</li>
 			<li>Brian</li>
 		</ul>
 	</my-panel>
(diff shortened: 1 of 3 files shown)


demos/can-component/accordion.html
```diff
@@ -75,19 +75,21 @@
 
 Component.extend({
 	tag: "my-accordion",
-	ViewModel: AccordionViewModel,
-	leakScope: true
```
chasenlehara: I’m not sure how `leakScope` is supposed to work in 4.0, but it was really buggy in this demo.

demos/can-component/accordion.html
```diff
@@ -106,9 +107,9 @@
 
 out.addEventListener("click", function(ev){
 	var el = ev.target;
-	var parent = el.parentNode;
+	var parent = canViewModel(el.parentNode);
 	if (el.nodeName === "BUTTON") {
-		parent.setAttribute("title", "Users");
+		parent.title = "Users";
```
chasenlehara: This demo is used in [can-component’s ViewModel docs](https://canjs.github.io/next/doc/can-component.prototype.ViewModel.html#Valuespassedfromattributes) to show how you can change a component’s element attributes to update the view model, but I don’t think this is supported in 4.0, so the demo now shows using `can-view-model` to get the component’s view model and set a property.

PR #3818, opened 2017-12-22
Debug guide

justinbmeyer:


docs/can-guides/experiment/technology/technology.md
```diff
@@ -0,0 +1,600 @@
...
+## Key-Value Observables
+
+The [can-define/map/map DefineMap] and
+[can-define/list/list DefineList] __Observables__ define the logic and state
+in your application. For example, if we wanted to model
+a simple counter, we can use [can-define/map/map DefineMap] as follows:
+
+```js
+import DefineMap from "can-define/map/map";
+var Counter = DefineMap.extend("Counter",{
```
chasenlehara: The mixture of `import` and `var` stand out; any reason not to use `const` and `let` in this document?

docs/can-guides/experiment/technology/technology.md
```diff
@@ -0,0 +1,600 @@
...
+```js
+import Component from "can-component";
+import stache from "can-stache";
+import DefineMap from "can-define/map/map";
+import route from "can-route";
+import "can-stache-route-helpers";
+
+Component.extend({
+    tag: "my-app",
+    autoMount: true,
```
chasenlehara: Great that we’ll be able to take these out. :D

docs/can-guides/experiment/technology/technology.md
justinbmeyer: `const` and `let` are for the weak minded.  I'm part of the hard-core`var` uproar!

PR #3828, opened 2018-01-04
using connectedCallback instead of inserted event in CTA Recipe

phillipskevin:


phillipskevin:
I'm just going to merge this and update the /next site once the tests pass, but if you have any comments @chasenlehara or @justinbmeyer I'll address them in a follow-up PR.

phillipskevin:
oh, right. no tests 😄 .

PR #3829, opened 2018-01-08
Revert "using DefineMap.extend for ViewModels since the default is SimpleMap now"

phillipskevin:
This reverts commit eb38511f1cf62cd5f55af0c44e8f5536bf8c8f4b.

### DIFF
--- a/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
+++ b/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
@@ -14,7 +14,7 @@ In this guide, you will learn how to:
 
 The final widget looks like:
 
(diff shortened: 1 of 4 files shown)


docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
```diff
@@ -514,33 +514,33 @@ We will do this by:
...
-    ViewModel: DefineMap.extend({
+    ViewModel: {
       map: "any"
-    })
+    }
   });
   ```
 
-  A ViewModel's [can-component/connectedCallback] can be used to know when the component's element is inserted into the document as follows:
+  A ViewModel's [can-component/connectedCallback] can be used to know when the component's eement is inserted into the document as follows:
```
chasenlehara: This should still be included.

PR #3834, opened 2018-01-13
Changes value to default

justinbmeyer:
Instead of `value`, uses `default`.


matthewp:
Might what to check if donejs needs this too on the APIs page. major is the branch.

phillipskevin:
@justinbmeyer this is passing now.

PR #3859, opened 2018-01-25
Update bit-docs-tag-demo so iframes are resized

m-mujica:


chasenlehara:
This has been done in master.

Issue #3781, opened 2017-12-04
CanJS 4.0 migration guide


James0x57:
### Changes to watch for

**Stache:**
* `%event` and other special symbols are now on `scope`
  - for example, `%event` is now `scope.event`
* can-view-scope doesn't do implicit scope walking, so change to use `../` and whatnot. Or use `{{scope.find('foo')}}`

**Routes:**
* `can.route(":page", { page: "home" });` is now `can.route("{page}", { page: "home" });`
* `can.route.ready();` is now `can.route.start();`
* `can.route()` is now `can.route.register()`

**Packages**
* `can-stache/helpers/route` is now `can-stache-route-helpers`

**Code/ViewModels**
* `can.batch.start()` is now `can.queues.batch.start()`
* `can.batch.stop()` is now `can.queues.batch.stop()`

matthewp:
`can-stache/helpers/route` is now `can-stache-route-helpers`.

matthewp:
can-view-scope doesn't do implicit scope walking, so change to use `../` and whatnot.

phillipskevin:
The `.each` method has been removed from `can-define/map/map`, `can-define/list/list`, `can-map`, and `can-list`.

You should use `.forEach` or `canReflect.each` instead.

chasenlehara:
`can-define` uses `can.keys` as the event name for when keys are dispatched instead of `__keys`

phillipskevin:
stache's `log` helper is now `console.log` (the other `console` methods are available also).

phillipskevin:
These will no longer automatically call `bar` as a function:
```
{{bar}}
```

```
{{#bar}}
  ...
{{else}}
  ...
{{/bar}}
```

You need to now call them manually (passing `scope.helperOptions` if you're using `options.fn` or `options.inverse`):
```
{{bar()}}
```

```
{{#bar(options.fn)}}
  ...
{{else}}
  ...
{{/bar}}
```

chasenlehara:
can-route-pushstate should now allow the browser to open a new tab/window when the user is holding down a modifier key (e.g. control, command, etc.)

phillipskevin:
`inserted` and `removed` events on components should be moved to `connectedCallback` and `disconnectedCallback`s on the component's ViewModel.

Or you should run this to make `inserted` and `removed` work:
```
import domEvents from "can-dom-events";
import domMutateDomEvents from "can-dom-mutate/dom-events";

[...]

PR #3860, opened 2018-01-25
Upgrade the migration guide based on feedback.

matthewp:
This upgrades the migration guide based on feedback provided in
https://github.com/canjs/canjs/issues/3781.

Also adds a section on the old stache symbols such as `%event`.


matthewp:
More feedback in https://github.com/canjs/canjs/issues/3781#issuecomment-360793619 and https://github.com/canjs/canjs/pull/3863

PR #3863, opened 2018-01-25
Add information about calling Stache Helpers as function

frank-dspeed:
Its importent to point out for stache helpers that they now get called as function else they return as toString() there is no warning for that at present


matthewp:
Thanks! I'll merge in the morning and expand on this a little.

matthewp:
I've cherry-picked your commit into https://github.com/canjs/canjs/pull/3860 and will be merged there. Thanks again!

Issue #3848, opened 2018-01-19
Leave the first argument to extend() out of the guides


chasenlehara:
It’s a little confusing to new users to see the first argument to `extend()` and not know what it is, if it needs to match your variable’s name, etc.

Note that we want to leave this in the main API docs, where we want to teach it as a best practice.

christopherjbaker:
I have the code to do this on my codemod, but due to the objection by @justinbmeyer, I did not include it when running. Does this need more discussion?

justinbmeyer:
This does need more discussion.  I think it's really important folks do it now that `canReflect.getName()` and so much of our debugging tools are improved by these names.

PR #3919, opened 2018-02-01
Cleanup guide code

### LINKED ISSUE #3846: Use const/let in sample code
It’d be nice for all of the code in our guides and core APIs to show const/let; extra points for having _all_ of our example code show const/let instead of var.

Note that this is not talking about the source code, where we intend to continue using what’s supported by the browsers we support.

christopherjbaker:
This fixes #3846 and #3848.

### DIFF
--- a/docs/can-guides/commitment/recipes/canvas-clock/1-setup.js
+++ b/docs/can-guides/commitment/recipes/canvas-clock/1-setup.js
@@ -1,23 +1,20 @@
-const ClockControlsVM = can.DefineMap.extend("ClockControlsVM", {
-  time: {
-    value({ resolve }) {
-      const intervalID = setInterval(() => {
-        resolve( new Date() );
-      },1000);
-
-      resolve( new Date() );
-
-      return () => clearInterval(intervalID);
-    }
-  }
-});
-
-can.Component.extend({
-  tag: "clock-controls",
-  ViewModel: ClockControlsVM,
-  view: can.stache(`
-    <p>{{time}}</p>
-    <digital-clock time:from="time"/>
-    <analog-clock time:from="time"/>
-  `)
-});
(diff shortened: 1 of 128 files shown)


chasenlehara:
What was your process for making all the const/let changes? Is this something we could easily do across the other packages too?

christopherjbaker:
Unfortunately, it was very manual. You cannot always use `const`, though in our case you can most of the time.. If we had eslint set up, that has an automatic ability to do this, but we aren't set up to do that at the moment (and I don't think it could handle js blocks in markdown).

christopherjbaker:
This now has the same treatment as https://github.com/canjs/can-define/pull/317 and similar, though also applied to the guide code itself. The debug names remain, to be revisited with #3848

docs/can-guides/commitment/recipes/canvas-clock/1-setup.js
```diff
@@ -1,23 +1,20 @@
...
+		}
+	}
+} );
+can.Component.extend( {
+	tag: "clock-controls",
+	ViewModel: ClockControlsVM,
+	view: can.stache( `
+<p>{{time}}</p>
+<digital-clock time:from="time"/>
+<analog-clock time:from="time"/>
```
chasenlehara: This indentation is incorrect.

docs/can-guides/commitment/recipes/canvas-clock/1-setup.js
```diff
@@ -1,23 +1,20 @@
...
-        resolve( new Date() );
-      },1000);
-
-      resolve( new Date() );
-
-      return () => clearInterval(intervalID);
-    }
-  }
-});
-
```
chasenlehara: New lines like this are significant in our guides because of the line highlighting we use to show what was added.

PR #3924, opened 2018-02-07
Added default codemod and changed wording to better explain calling f…

Sinjhin:
…unctions with parens


matthewp:
Thanks!

phillipskevin:
I still think this might be confusing for people. I think we should just say

> functions should be called using parentheses

Then show what this means and give the explanation of why we did this (like you are).

Then once this is clear, say something like

> there is one exception -- the Helper Expression syntax, which is allowed when your function is called with at least one argument. This is for legacy purposes, so you don't have to change all of the places you're doing `{{#each list}}` in your app.

phillipskevin:
My main point is, functions _should_ be called with `()`. There's one situation where they don't _have_ to be.

PR #3927, opened 2018-02-08
changed wording based on comments

Sinjhin:


docs/can-guides/upgrade/migrating-to-canjs-4.md
```diff
@@ -98,22 +98,33 @@ The first step to upgrading to CanJS 4 is to deal with the breaking changes. Mos
...
+
+This is to prevent syntax ambiguity.
+
+Consider:
 ```
 {{foo}}
 ```
 
-Is `foo` a function or a value? If `foo` is not a built-in helper and has no arguments, ex. `{{foo argument}}`, then we treat it as a value unless `()` are present.
+Is `foo` a function or a value? If `foo` is not a built-in helper and has no arguments, ex. `{{foo argument}}`, then we treat it as a value and use the `toString` to resolve the value unless `()` are present.
```
phillipskevin: > It's impossible to tell when reading this that `foo` might be a function that will be called.

> The only exception is built-in helpers or Helper Expressions (when called with >=1 argument)

docs/can-guides/upgrade/migrating-to-canjs-4.md
```diff
@@ -98,22 +98,33 @@ The first step to upgrading to CanJS 4 is to deal with the breaking changes. Mos
...
 
-This is true of both methods on the ViewModel and helpers with no arguments. It does not apply to built-in helpers or registered helpers with arguments.
+This is true of both methods on the ViewModel and helpers with no arguments. It does not apply to built-in helpers (such as each, eq, if, etc...) or registered helpers with at least one argument.
 
 ```handlebars
```
phillipskevin: double check how this looks when rendered.

PR #3987, opened 2018-02-23
update bit-docs dependencies for Prism

christopherjbaker:


christopherjbaker:
Current Issues:
* Shadow under code blocks
* Alternating line colors drift
* Horizontal scrolling?
* Some newlines being removed? (/doc/guides/todomvc.html)
* [Highlight](bit-docs/bit-docs-html-highlight-line) doesn't look good

PR #3999, opened 2018-02-28
Add can-map-define

m-mujica:


package.json
```diff
@@ -70,7 +70,8 @@
     "can-key-tree": "1.0.0",
     "can-list": "4.0.0",
     "can-log": "1.0.0",
-    "can-map": "4.0.0",
+    "can-map": "4.0.1",
+    "can-map-define": "^4.0.0",
```
phillipskevin: can you pin the version?

PR #4047, opened 2018-03-15
Minor site improvements

chasenlehara:
- Update route() calls to route.register() in some docs & demos
- Replace var with const/let in a few places

### DIFF
--- a/docs/can-canjs/canjs.md
+++ b/docs/can-canjs/canjs.md
@@ -50,7 +50,7 @@ by extending [can-component] with:
   my-counter button {margin-left: 15px;}
 </style>
 <script type="text/steal-module">
-var Component = require("can-component");
+const Component = require("can-component");
 Component.extend({
     tag: "my-counter",
     view: `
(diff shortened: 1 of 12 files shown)


docs/can-canjs/canjs.md
```diff
@@ -50,7 +50,7 @@ by extending [can-component] with:
   my-counter button {margin-left: 15px;}
 </style>
 <script type="text/steal-module">
-var Component = require("can-component");
+const Component = require("can-component");
```
chasenlehara: I think this should be an import statement, but babel wasn’t loading for me when I tested the site locally, so I didn’t want to mess with it.

docs/can-guides/upgrade/using-codemods.md
```diff
@@ -350,13 +350,13 @@ can-migrate -a **/*.js -t version-4/can-route/page.js
 …will transform the following:
 
 ```js
-can.route(":page", { page: "home" });
+can.route.register(":page", { page: "home" });
```
phillipskevin: The transform actually handles both since it might run before the `register` transform, but this is fine.

chasenlehara: That was sloppy on my part, I’ve restored how this guide was before.

PR #4054, opened 2018-03-19
Minor edits to the Forms guide

chasenlehara:
- Add a section to the end to prompt people to ask questions on our forums or Gitter
- Wrap form elements in labels where appropriate
- Other minor changes

### DIFF
--- a/docs/can-guides/topics/forms.md
+++ b/docs/can-guides/topics/forms.md
@@ -17,7 +17,7 @@ There are three main tasks when working with forms and form elements:
 CanJS provides many useful utilities and techniques to help with these tasks. This guide will walk you through:
 
 * binding to form elements using [guides/forms#Attributebinding one-way], [guides/forms#Two_waybinding two-way], and [guides/forms#Eventbinding event] binding
-* combining bindings in a [guides/forms#Datadown_actionsup "data down, actions up"] pattern
+* combining bindings in a [guides/forms#Datadown_actionsup “data down, actions up”] pattern
 * using shorthand bindings like [guides/forms#Bindingattributesonspecificevents `on:input:value:to`]
 * storing view-related data in [guides/forms#Bindingtotemplatevariables template variables]
 * working with [guides/forms#Commonformelements common form elements]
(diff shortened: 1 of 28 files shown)


docs/can-guides/topics/forms.md
```diff
@@ -29,17 +29,17 @@ CanJS provides many useful utilities and techniques to help with these tasks. Th
...
 
-The following sections will explain how to use each of these bindings on their own as well as how to combine them in more advanced binding formats.
+The following sections will explain how to use each of these bindings on their own, as well as how to combine them in more advanced binding formats.
 
 ### Event binding
 
-Event bindings allow you to respond to DOM Events. To set up an event binding, add an attribute to your element like `on:DOM_EVENT="CALL_EXPRESSION"` where `DOM_EVENT` is the event you want to respond to and `CALL_EXPRESSION` is a [can-stache/expressions/call Call Expression] for the function you want to call when the event occurs. Here is a simple example:
+Event bindings allow you to respond to [https://developer.mozilla.org/en-US/docs/Web/Events DOM Events]. To set up an event binding, add an attribute to your element like `on:DOM_EVENT="CALL_EXPRESSION"` where `DOM_EVENT` is the event you want to respond to and `CALL_EXPRESSION` is a [can-stache/expressions/call Call Expression] for the function you want to call when the event occurs. Here is a simple example:
 
 @demo demos/forms/bindings-event.html
```
chasenlehara: This demo uses `innerHTML` instead of having a `count` property on a view-model and rendering it with `{{count}}`. That stuck out to me and I wonder if it’ll be jarring to other people too.

phillipskevin: I did it this way to use as little stache as possible, so that the event binding was the only thing being introduced. You're right, that might make it more confusing. I'm open to suggestions.

docs/can-guides/topics/forms.md
```diff
@@ -490,252 +492,252 @@ There are many useful techniques for working with related data. Before diving in
...
 The `makes` property is set to the result of calling [can-ajax], which returns a [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise promise].
 
-This promise is decorated by [can-reflect-promise] so that in the view you can easily use metadata related to the promise's [https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md#states state]:
+This promise is decorated by [can-reflect-promise] so that in the view you can easily use metadata related to the promise’s [https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md#states state]:
 
-* `state` - one of "pending", "resolved", or "rejected"
-* `isPending` - true if the promise is neither resolved nor rejected, false otherwise.
-* `isResolved` - true if the promise is resolved, false otherwise.
-* `isRejected` - true if the promise is rejected, false otherwise.
+* `state` — one of `pending`, `resolved`, or `rejected`
```
phillipskevin: Since these are strings, maybe it should be:
> `"pending"`, `"resolved"`, or `"rejected"`

chasenlehara: I don’t think anyone would be confused if those strings weren’t wrapped in quotation marks, but for the sake of being consistent with can-reflect-promise, I’ve reverted back to what you had.

docs/can-guides/topics/forms.md
chasenlehara: Yeah, that makes sense. I’ve updated the example to use `{{count}}`; it’s about the same amount of code and I think it makes it more consistent with the other examples.

PR #4085, opened 2018-03-30
Adds the can/es module for tree-shaking bundlers

### LINKED ISSUE #4013: Make CanJS tree-shakable
__TLDR:__ Make it so the `can` package export is tree-shake-able. Allowing you to import just what you want like:

```js
import {Component, observe} from "can"
```

> This was discussed on a [recent live stream (8:14)](https://youtu.be/EEBytcIX7oA?t=8m14s).

## The Problem 

Currently, to use CanJS, you likely have to install at least 8 different packages:

```
npm install can-component can-connect can-define can-route can-route-pushstate can-set can-stache can-stache-bindings --save
```

While this is for good reason, it's annoying.

Furthermore, to build something like a model, you need to import a bunch of stuff:

```js
var DefineMap = require("can-define/map/map");
var DefineList = require("can-define/list/list");
var superMap = require("can-connect/can/super-map/super-map")
var set = require("can-set");
```

## The Solution

I'd like to make the `can` package's main export a tree-shake-able `ES` module.  This would allow importing enough to make a model look like:

```js
import {DefineMap, DefineList, superMap, set} from "can";
```

This module should be tree-shake-able by both webpack and stealjs.  So if someone imports only, for example `ajax` like:

```js
import {ajax} from "can";
```

They would only get `ajax` in their build.

### What is the minimal viable example of this?

Ideally, we will not have to rewrite every existing project to `ES` modules.  Instead, `steal.export()` should be able to do the necessary transformations.

Also, it would probably be nice if we don't have to move every module within the single export.  The main module might look like:

```js
// can.js 
exports Component from "can-component";
exports route from "can-route";
```

Essentially, we will export everything by it's `can-namespace` name.  Perhaps there's a way to do this programmatically within the module, but doubt it.  Something like:

```js
import "moduleA";
import "moduleB";
...
import "moduleZ";
import can from "can-namespace";
for(var prop in can){
  exports [prop] as can[prop]
}
```

But I would be fine maintaining this list of modules and exports for now.

### How do we document this?

There are a few places in the docs that might be affected by using tree-shaking as the primary way of using CanJS:

- The sidebar
- The title
- Examples

I still think CanJS's sidebar should use the `module-name` listing.  But we should show the alias, perhaps next to the module name in the title area like:

canjs_-_can-component

What should examples how?

Guides should almost certainly show using the `tree-shaking` module. This will make JSBin / script tag use very similar.  But what about individual packages and modules?

matthewp:
This change adds a new module, `can/es` which exports everything from
__core__ as named exports. This enables people to import only the parts
of CanJS they need:

```js
import { Component, DefineMap } from "can/es";
```

Provided they use a tree-shaking capable bundler, the bundle size will
not be too large.

Closes #4013

### DIFF
--- a/docs/can-guides/experiment/advanced-setting-up.md
+++ b/docs/can-guides/experiment/advanced-setting-up.md
@@ -1,0 +1,48 @@
+@page guides/advanced-setup Experimental ES Module Usage
+@parent guides/getting-started 4
+@outline 2
+
+@description Use [ES modules](http://exploringjs.com/es6/ch_modules.html)' *named exports* feature to import just the APIs of CanJS that you are using.
+
+@body
+
+CanJS provides an ES module that contains [named exports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#Using_named_exports) which can be imported and used to pluck out the parts that you need. When used in conjunction with [tree shaking](http://exploringjs.com/es6/ch_modules.html#_benefit-dead-code-elimination-during-bundling) you gain:
+
+* Fewer packages to import in each of your modules.
+* Bundles that exclude all of the parts of CanJS that you don't use.
+
+To use, import the `can/es` module like so:
+
+```js
+import { Component, DefineMap } from "can/es";
+
+const ViewModel = DefineMap.extend({
+	message: "string"
+});
+
+Component.extend({
+	tag: "my-component",
+	ViewModel
+});
+```
+
+If you are using [webpack](https://webpack.js.org/guides/tree-shaking/#src/components/Sidebar/Sidebar.jsx) tree-shaking is only available when in production mode. The following config shows a typical setup:
+
+__webpack.config.js__
+
+```js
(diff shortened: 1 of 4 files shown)


docs/can-guides/experiment/advanced-setting-up.md
```diff
@@ -0,0 +1,48 @@
...
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname)
+  }
+};
+```
+
+## Names
+
+The names that can be exported by this module mirror what is part of the `can` namespace object that you get from `import can from "can"`. You can see the names that `can/es` exports [here](https://github.com/canjs/canjs/blob/master/es.js).
```
phillipskevin: should this be
> The names that can be imported from this module...

or 

> The names that are exported by this module...

matthewp: I'll go with the former.

PR #4097, opened 2018-04-06
Remove mention of IE in our browser support list

chasenlehara:
Also bump the iOS Safari version to what Sauce Labs tests.

### DIFF
--- a/docs/can-guides/introduction/technical.md
+++ b/docs/can-guides/introduction/technical.md
@@ -11,8 +11,7 @@
 CanJS supports:
 
  - Latest Chrome, Firefox, Edge, Safari
- - IE9+
- - iOS Safari 10+
+ - iOS Safari 11+
  - NodeJS 6+ with [can-vdom] as a document.
 
 [![Sauce Test Status](https://saucelabs.com/browser-matrix/canjs.svg)](https://saucelabs.com/u/canjs)


justinbmeyer:
CanJS 3.0 supports IE and will continue to. 3.0 has been out for a year and a half. 

Or plan for 4.0 is to support IE11 via polyfills. I think someone in the community has been working on this. I’m on my phone so too lazy to find the issue. 

But we didn’t want it to seem like IE 11 is supported right now, so we removed it until we get all tests passing with the polyfills. 

Edge is supported in 4.0.

rjgotten:
> CanJS 3.0 supports IE and will continue to. 3.0 has been out for a year and a half. 

Doesn't add enough value over the private fork of 2.x that I have in use. I already back-ported the most-desired features we saw in 3.0 like `<can-slot>` to said fork. 4.x did bring interesting improvements performance- and stability-wise (queue-based event dispatches, etc) and cleaned up a lot of mess with binding and scope handling in stache.

> Or plan for 4.0 is to support IE11 via polyfills.

If you're in full control of the site or app that is meant to run your CanJS-based logic, that's a solution.
But it's infeasible when you're not. E.g. when you're active as a third-party being integrated on an existing site.

justinbmeyer:
Regarding the polyfill approach, we required the same thing for IE 6 and 7 for old versions of CanJS.  You had to include HTML5Shiv. We also require a Promise polyfill for canjs 3.0. 

Polyfills are a pretty common practice. Why are you hostile to them?

justinbmeyer:
> If you're in full control of the site or app that is meant to run your CanJS-based logic, that's a solution.
But it's infeasible when you're not. 

I'd like to hear more about this situation.  To me, it seems like the solution is to use polyfills that feature detect the absence of the desired feature and only polyfill it when it is missing.  Would this approach work for you?

Alternately, we could have everything in CanJS import these features like:

```js
var WeakMap = require("can-polyfill/weakmap");
```

Which would look like this:

```js
// can-polyfill/weakmap.js
module.exports = WeakMap
```

Then, folks using a module loader could map from `can-polyfill/weakmap` to an actual polyfill.  The benefit of this approach is that WeakMap wouldn't be polluted.


===========

Anyhoo, all of the above is to say, is that we will try to work with people to find solutions to upgrading to 4.0.  Please detail more of your concerns in that issue: #4098.

rjgotten:
> Polyfills are a pretty common practice. Why are you hostile to them?

Because polyfills have to modify the global environment, which is a big no-no if you have to produce code that should be capable of running as a third-party, integrated into an existing website by others.

Having to polyfill Promise, or ask that integrators have it polyfilled, isn't that bad. Not very different from how back in the day lots of third-party integrations would warn you that they would load jQuery or could take advantage of a copy of jQuery already loaded.

But having to set up polyfills for a bevy of other things like `Object.assign`; `WeakMap`; `WeakSet`; etc. becomes too much for that to be managable.

Contrast with an implementation that imports everything as a module and falls through to the native implementation _when available_ and otherwise falls back to a non-global polyfilled fallback.

E.g. for `Object.assign`:

```js
module.exports = ( "assign" in Object.prototype ) 
  ? function assign( object, props ) {
    return Object.assign( object, props ); 
  } : function assign( object, props ) {
    // FALLBACK IMPLEMENTATION
  }
```

justinbmeyer:
@rjgotten I'm ok with removing a dependency on `Object.assign` because we already have `canReflect.assignMap()`. I'm not sure how that actually snuck in.  But WeakMap and WeakSet were one, if not __THE__, critical design choice that helped simplify CanJS and make a lot of features either possible, or manageable. 

I don't want to burden people with pulling in polyfills as CanJS's default behavior.  

I'd like some additional clarity on the root problem:

- Is it the number of polyfill "files"? 
- Is it the number of polyfilled "APIs"?
- Is it some mixture of both?

I ask this because presumably you don't have to ask people to setup any polyfill "files". They can be built into your app file.

If it's the number of "APIs", then how many is too much?  

Would the module-loader mapping technique work for you?

PR #4127, opened 2018-05-03
Includes docs for can-query and such

justinbmeyer:


justinbmeyer:
This is breaking in Edge 17.  I'm not able to launch it with browserstack.  But everything passes in Edge 16 so I'm going to merge.

PR #4140, opened 2018-05-22
Update API Docs page with current CanJS 4 packages

chasenlehara:
This removes any mention of packages that haven’t been updated for CanJS 4 and updates some others (like moving can-compute to legacy).

### DIFF
--- a/docs/can-canjs/can-api.md
+++ b/docs/can-canjs/can-api.md
@@ -81,6 +81,7 @@ __Core Collection__
   "can-set": "<%can-set.package.version%>",
   "can-stache": "<%can-stache.package.version%>",
   "can-stache-bindings": "<%can-stache-bindings.package.version%>",
+  "can-stache-route-helpers": "<%can-stache-route-helpers.package.version%>",
 ```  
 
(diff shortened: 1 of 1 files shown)


docs/can-canjs/can-api.md
```diff
@@ -283,21 +272,9 @@ _Useful libraries that add important features or extend the core collection._
 - **[can-fixture]** <small><%can-fixture.package.version%></small> Simulate AJAX requests
   - `npm install can-fixture --save`
   - <a class="github-button" href="https://github.com/canjs/can-fixture">Star</a>
-- **[can-fixture-socket]** <small><%can-fixture-socket.package.version%></small> Simulate sockets
```
phillipskevin: I think `can-fixture-socket` should still be compatible with 4.0. Its dependencies are the same as `canjs/canjs#master`.

chasenlehara: It’s not included in the build or site right now, so I’m going to remove it and see if it can be added back in with https://github.com/canjs/canjs/issues/4142

PR #4192, opened 2018-06-12
Export routePushstate from can.js

matthewp:


justinbmeyer:
Will this turn on pushstate routing in all jsbins and such?

Maybe in 5.0, we should make people call `routePushstate()` or something ...

matthewp:
I guess it might... what other side-effectual modules are there that we don't want to be turned on by default?

matthewp:
I guess we shouldn't do this.

justinbmeyer:
I think we should, but change `can-route-pushstate()` so you have to call it ... though this will effect done-ssr.

matthewp:
@justinbmeyer let's discuss in can-route-pushstate if you are proposing changing its API. I'm not sure what you mean by calling it... it returns an object currently.

PR #4199, opened 2018-06-15
Fix demos to remove `~` and fix issue with inList demo checkboxes not…

Mattchewone:
… working

Fixes: https://github.com/canjs/can-stache-converters/issues/79


matthewp:
Thanks! Why is `~` not needed now?

justinbmeyer:
@matthewp when you register a converter, it will always call the converter with observables.

PR #4220, opened 2018-06-27
Upgrade to can-component, can-stache-bindings pre

matthewp:
This upgrades the two packages to their prerelease. This is based on
can-stache-bindings no longer registering its own attr callbacks, but
instead defering to `stache.addBindings()` which is done by
can-component.


matthewp:
We're going to make can-stache-bindings be non-breaking.

PR #4234, opened 2018-06-29
Use distinctive identifier names

CODE-REaD:
Changed a use of `ViewModel` to `selectVM` and a use of `items` to `titleItems` to distinguish them from keywords and other identifiers with the same name.  See https://forums.donejs.com/t/generic-property-names-confuse-the-beginner/908 for reasons why.

### DIFF
--- a/demos/can-component/selected.html
+++ b/demos/can-component/selected.html
@@ -24,7 +24,7 @@ var DefineMap = require("can-define/map/map");
 var DefineList = require("can-define/list/list");
 var stache = require("can-stache");
 
-var ViewModel = DefineMap.extend({
+var selectVM = DefineMap.extend({
 	selected: {
 		Default: DefineList
 	},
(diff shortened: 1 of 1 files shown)


demos/can-component/selected.html
```diff
@@ -24,7 +24,7 @@
 var DefineList = require("can-define/list/list");
 var stache = require("can-stache");
 
-var ViewModel = DefineMap.extend({
+var SelectViewModel = DefineMap.extend("SelectViewModel", {
```
chasenlehara: Constructor functions are usually capitalized, which is why I changed it from `selectVM` to `SelectViewModel`.

Also, we don’t show this everywhere, but if you provide a string as the first argument to `.extend()`, it’ll have a nicer name in the debugger. https://canjs.com/doc/can-define/map/map.extend.html

chasenlehara:
Thanks @CODE-REaD!

PR #4277, opened 2018-07-05
adding doc folder to npmignore

phillipskevin:


phillipskevin:
just going to do this on major since CI takes forever.

PR #4290, opened 2018-07-06
WIP: speeding up CI

phillipskevin:


matthewp:
Nice! 👍

phillipskevin:
Going to go with this for now. It takes the CI time down to about 33 minutes. I'll put in a follow-up PR once I hear back from Travis Support about only adding the sauce_connect addon for the one `env` that needs it.

Issue #4116, opened 2018-04-20
Improve the content of the CanJS documentation


chasenlehara:
__TLDR:__ we propose reviewing our current docs to look for:

- pages that don’t have useful content
- broken links
- bad/broken code examples & demos
- other ways we can improve them to make them more useful

## Problem

Our docs have evolved over the years without a thorough review. Inevitably, some docs have deteriorated with broken links, out-of-date examples, and busted demos. Some pages lack good content or examples.

## Solution

Review the most popular pages (maybe top 50?) to:

- Fix broken links, code examples, demos, etc.
- Look for ways to improve the content

Additionally, we should tackle these existing issues:

- [Consistently use const/let in sample code](https://github.com/canjs/canjs/issues/3846)
- [Consistently use ES6 import syntax instead of CJS in sample code](https://github.com/canjs/canjs/issues/3847)
- [Improve the descriptions of (at least) the Core packages](https://github.com/canjs/canjs/issues/4057)
- [Improve docs on loading templates](https://github.com/canjs/can-stache/issues/429)

We should also look at [issues with the “documentation” label](https://github.com/search?o=desc&q=org%3Acanjs+is%3Aopen+is%3Aissue+label%3Adocumentation&s=created&type=Issues).

As a stretch goal, we should consider:

- [Adding a way for users to give us feedback on whether docs are good/useful](https://github.com/canjs/bit-docs-html-canjs/issues/451)
- Making it possible for our demos/examples to open in an online code editor

chasenlehara:
This isn’t something we have to address as part of tackling this proposal, but where we’re already updating code… @code-read brings up a good point that more distinctive variable & property names can make it easier to understand code examples: https://forums.donejs.com/t/generic-property-names-confuse-the-beginner/908

justinbmeyer:
## Priority Guides 

- [ ] Setting Up CanJS 
- [ ] Technology Overview 
- [ ] HTML 
- [ ] Routing 
- [ ] Service Layer 
- [ ] Debugging (Kevin)


## Core 

- [ ] can-component
- [ ] can-define
- [ ] can-stache
- [ ] can-stache-bindings
- [ ] can-route
- [ ] can-value
- [ ] can-route (Kevin)
- [ ] can-route-pushstate (Kevin)
- [ ] can-stache-route-helpers (Kevin)
- [ ] can-fixture 
- [ ] can-query-logic 
- [ ] can-realtime-rest-model 
- [ ] can-rest-model 


[...]

PR #4317, opened 2018-07-12
Replaces var with let/const in the canjs/docs files. Fixes #3846 & addresses #4116

### LINKED ISSUE #3846: Use const/let in sample code

Aosanders:


christopherjbaker:
Might be worth running this codemod against all the markdown files in this repo. It looks like it didn't get run against `canjs/canjs` when i ran it against everything else.

PR #4372, opened 2018-07-26
Remove unused get componentToShow() definition

CODE-REaD:


docs/can-guides/commitment/recipes/modals/3-multiple-modals.js
```diff
@@ -103,9 +103,6 @@ Component.extend({
 		{{/ each}}
 	`,
 	ViewModel: {
-		get componentToShow() {
```
justinbmeyer: Thanks! I wonder if this affects the line numbers.

PR #4374, opened 2018-07-27
Demo fixes

mikedane:
Fixes broken demos not working when opening in codepen using `edit in codepen` button


michaelzcheng:
Fixed demos for codepen. Broken demos in codepen on can-component and it's viewModel sub-section should be fixed by the new canjs release. Will QA upon release.

demos/can-component/accordion.html
```diff
@@ -14,23 +17,14 @@
...
+</my-accordion>`;
 
-var Panel = DefineMap.extend({
+const Panel = DefineMap.extend({
 	active: "boolean",
 	title: "string"
 });
 
-var AccordionViewModel = DefineMap.extend({
+const AccordionViewModel = DefineMap.extend({
```
justinbmeyer: We should probably change these eventually to the "standard" component syntax, without everything broken out.

PR #4425, opened 2018-08-27
load ecosystem from everything.mjs

justinbmeyer:
This changes the build and rewrites the getting started guide to be consistent with how to load ecosystem modules.  It also includes an eye on how to make this work with a codepen.

In short, we will have a `can/everything.js` which is a copy of `can.js`. This is only indented to be a signal that we are loading some ecosystem modules.  Ideally users would not see `can/everything.js`.  Though if they do, that's ok for now an loading `can` or `can/everything.js` won't really matter for their app.

When we make a codepen, we will change `can/everything.js` to `unpkg.com/can@5/everything.mjs`.


phillipskevin:
🎉

PR #4544, opened 2018-10-20
Update routing.md

leoj3n:
Improve the "Routing and the root component" section.

### DIFF
--- a/docs/can-guides/experiment/technology/routing.md
+++ b/docs/can-guides/experiment/technology/routing.md
@@ -217,9 +217,9 @@ building `<my-app>` and connecting it
 to [can-route]. This is usually done in five steps:
 
 1. Define the top-level component’s view-model (sometimes called _application view-model_).
-2. Create an observable key-value object on the view-model to represent the state of [can-route].
+2. Create an observable key-value object (a [can-define/map/map DefineMap]) on the view-model to represent the state of [can-route].
 3. Connect this observable to the routing [can-route.data].
-4. Have the top-level component’s [can-component.prototype.view] display the corrent sub-components based on the view-model state.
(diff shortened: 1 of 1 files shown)


leoj3n:
Actually, I don't think the bit I added about `DefineMap` is correct?

leoj3n:
Odd question:

Is the correct terminology "on the view-model" or "in the view-model"?

Or are "in" and "on" interchangeable when talking about the view-model DefineMap?

docs/can-guides/experiment/technology/routing.md
```diff
@@ -216,10 +216,10 @@ In most applications, [can-route] is connected to a property on the top-level co
 building `<my-app>` and connecting it
 to [can-route]. This is usually done in five steps:
 
-1. Define the top-level component’s view-model (sometimes called _application view-model_).
+1. Define the top-level component’s view-model (a [can-define/map/map DefineMap] sometimes called _application view-model_).
```
chasenlehara: The VM can be a can-map, or can-observe, or…

Was there something confusing about not mentioning the type here?

matthewp: I think it's always good to add links though, don't you think? I think the wording could be changed to be more non-specific.  *such as a DefineMap*

leoj3n: Is it most likely to be a _DefineMap_? I don't think there's any harm in pointing out what it's most likely going to be. If so, instead of _such as a DefineMap_, it could read _typically a DefineMap_. Otherwise should probably say _an Object_ or just not change this sentence.

chasenlehara: I made a small edit to say “typically.” Thanks!

PR #4605, opened 2018-11-20
Update the “Signup and Login” guide for CanJS 5

chasenlehara:
I took some liberties in updating this guide. This includes:

- Switching from showing `can.stache.from` to using a real component
- Using the new simple setter syntax (e.g. `on:click="this.page = 'login'"`)
- Using `this.` throughout the template
- Completely moving the template to a JS string literal; now the HTML only has the custom element
- Moving the fixture setup into the JS


docs/can-guides/commitment/recipes/signup-simple/6-login-errors.js
```diff
@@ -1,68 +1,152 @@
...
+	          <a href="javascript://" on:click="this.page = 'signup'">Sign up</a>
+	        </aside>
+	      </form>
+
+	    {{/ eq }}
+
+	  {{/ if }}
+	`,
+	ViewModel: {
+    sessionPromise: {
```
phillipskevin: There's some weird spacing here.

phillipskevin: looks good.

PR #4666, opened 2018-12-12
Update can-define.js

frank-dspeed:
can-define needs still a default export because of EMCA Loader Specs else it gets harder to require it

### DIFF
--- a/es/can-define.js
+++ b/es/can-define.js
@@ -1,3 +1,3 @@
-export { default as define } from "can-define";
+export { default as define, default } from "can-define";
 export { default as DefineMap } from "can-define/map/map";
 export { default as DefineList } from "can-define/list/list";


matthewp:
es/can-define is not meant to be used directly. Is that what you are trying to do?

frank-dspeed:
@matthewp not really i was trying to play around with some converter tools and found out that it makes bad results 

i wanted to slim canjs a bit as it got tons of doubled functions and exports and all that and i get always a 1.3 megabyte unminifyed version of core.mjs this was much to big compared to any other lib 

i am trying to get a canjs that really only loads what is used and found out on my research that core.mjs is not as treeshake compatible as expected 

the only good way i found is making it even more modular or replace a lot of stuff with dynamic imports

but at all i come to the final result i should forget about the original codebase complet and use only the canjs concepts and refactor it complet by hand 


- 1.3M    src/ is my bundle result

```
import DefineMap from 'can-define/map/map'
import Component from 'can-component'
console.log(DefineMap,Component)
```

matthewp:
What problems did you find with tree-shaking core.mjs?

frank-dspeed:
@matthewp it simply gets not treeshaked it got complet  importet and i don't know why exactly we are debuging that but i think its { } from Syntax that makes the fail we plan to deprecate that all over

we know react and all are using it but it makes to much trouble but i think we can close this as your right this should not get used directly 

i think i will jump a bit into hyperHTML with Neverland ( A succesor over your haunted Package)


https://github.com/WebReflection/neverland

PR #4699, opened 2019-01-05
Inital Changes for es-modules branch

frank-dspeed:
Please don't merge will test tomorrow if all is in shape and then update this PR

it Includes at present only the basic needed changes.

It will produce highly optimized es-modules for every can-* package under the /es directly inside of dist/es/can-*.mjs
that are directly use able via 

```js
import Connect from 'https://unpkg.com/canjs@5.16.1/dist/can-connect.mjs'
```

it is a core part for this https://github.com/donejs/donejs/issues/1165 a more detailed roadmap for it exists under https://github.com/donejs/donejs/issues/1167


chasenlehara:
@frank-dspeed Could you please outline what this PR accomplishes?

Issue #4770, opened 2019-03-14
Should we export can-kefir as { kefir } or { Kefir } ?


chasenlehara:
Kefir’s docs show uppercase `Kefir`: https://kefirjs.github.io/kefir/

can-kefir’s docs follow suit and show uppercase `Kefir`: https://canjs.com/doc/can-kefir.html

…but as far as I can tell, Kefir isn’t a constructor function. I don’t see any reason why it should be uppercase, other than to be consistent with the original Kefir docs.

Right now, we export lowercase `kefir`: https://github.com/canjs/canjs/blob/7b6aed59b9040d4f6b28ca43cf46755df5c60744/ecosystem.js#L6

Should we keep this and update our docs to show lowercase `kefir`? Or should we export `Kefir` instead?

This blocks [updating the advanced credit card guide](https://github.com/canjs/canjs/issues/4622).

phillipskevin:
I think we should keep it consistent with Kefir's docs and Kefir's code and export `Kefir`.

PR #4796, opened 2019-04-05
Fix export Kefir 

cherifGsoul:
With this `Kefir` is exported instead of `kefir`.
for #4770


ecosystem.js
```diff
@@ -3,7 +3,7 @@ export { default as debug } from "./es/can-debug";
 export { default as defineBackup } from "./es/can-define-backup";
 export { default as defineStream } from "./es/can-define-stream";
 export { default as defineStreamKefir } from "./es/can-define-stream-kefir";
-export { default as kefir } from "./es/can-kefir";
+export { default as Kefir } from "./es/can-kefir";
```
chasenlehara: This is a breaking change. We should export both for now.

cherifGsoul: Fixed

chasenlehara: Does this work?

```js
export { default as kefir, default as Kefir } from "./es/can-kefir";
```

cherifGsoul: I'll check it.

Issue #4340, opened 2018-07-20
Potentially misleading build instructions on "Improving the Docs & Site" page


mikedane:
https://canjs.com/doc/guides/contributing/documentation.html#Improvethetheme_sdesignandstyles

Going through the above linked page on the contribution guide, I noticed a potentially misleading set of statements

<img width="1380" alt="screen shot 2018-07-20 at 8 31 11 am" src="https://user-images.githubusercontent.com/13106307/43002708-bda8b166-8bf7-11e8-95c2-13ba40919dba.png">

Highlighted in the above image it's instructing the user to use `npm run document:force` to build the docs, however once that command has been run once, the user can simply use the less time consuming `npm run document` command without the `:force` modifier. This should be noted somewhere in this section as it saves the user a ton of time when testing documentation changes

matthewp:
Are you sure that's true? When you install a new package you might need to run force again. cc @justinbmeyer who'd know for sure.

mikedane:
Yeah good point! Im definitely not 100% sure, although I was using this yesterday and was able to successfully run `npm install ../can-compute && npm run document` which is technically installing a new package. Either way I think we should put some sort of note here which better explains the correct usage. @justinbmeyer

PR #4834, opened 2019-04-25
Remove `:force` from the `npm run document` command

Mattchewone:
Only show the `npm run document` script as the `:force` version can be misleading and slow.

Ref - https://github.com/canjs/canjs/issues/4340


chasenlehara:
What’s misleading about using `:force`?

Mattchewone:
In https://github.com/canjs/canjs/issues/4340 @mikedane mentions about `:force` not being required, and @justinbmeyer also mentions that unless `bit-docs` config changes running `npm run document` will be sufficient.

PR #4878, opened 2019-05-13
can-stache transforms default targets

mikemitchel:
`can-stache` transform examples should use `.stache` in the glob by default as that is a more likely target than an inline string literal template in a `.js` file.

### DIFF
--- a/docs/can-guides/upgrade/using-codemods.md
+++ b/docs/can-guides/upgrade/using-codemods.md
@@ -91,10 +91,10 @@ Runs all the `can-migrate` transforms for upgrading to CanJS 4.0 on the files th
 can-migrate '**/*.js' --can-version 4 --apply
 ```
 
-Runs the `can-stache/route-helpers` transform on the files that match the `**/*.js` glob:
+Runs the `can-stache/route-helpers` transform on the files that match the `**/*.stache` glob:
 
 ```bash
-can-migrate '**/*.js' --transform can-stache/route-helpers.js --apply
+can-migrate '**/*.stache' --transform can-stache/route-helpers.js --apply
 ```
 
 You can find a [complete list of version-4 transforms on GitHub](https://github.com/canjs/can-migrate/tree/master/src/transforms/version-4).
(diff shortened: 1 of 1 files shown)


docs/can-guides/upgrade/using-codemods.md
```diff
@@ -91,10 +91,10 @@ Runs all the `can-migrate` transforms for upgrading to CanJS 4.0 on the files th
 can-migrate '**/*.js' --can-version 4 --apply
 ```
 
-Runs the `can-stache/route-helpers` transform on the files that match the `**/*.js` glob:
+Runs the `can-stache/route-helpers` transform on the files that match the `**/*.stache` glob:
 
 ```bash
-can-migrate '**/*.js' --transform can-stache/route-helpers.js --apply
+can-migrate '**/*.stache' --transform can-stache/route-helpers.js --apply
```
chasenlehara: I think this one’s confusing because it applies to both JS & stache. I’d leave it as-is or change the example to something that only applies to one or the other.

docs/can-guides/upgrade/using-codemods.md
```diff
@@ -306,15 +306,15 @@ can.route.start();
 To run [all of the version-4/can-stache transforms](https://github.com/canjs/can-migrate/tree/master/src/transforms/version-4/can-stache) listed below:
 
 ```bash
-can-migrate '**/*.js' -t version-4/can-stache/ -a
+can-migrate '**/*.stache' -t version-4/can-stache/ -a
```
chasenlehara: I think we should show both because the transforms apply to both types of files.

docs/can-guides/upgrade/using-codemods.md
```diff
@@ -404,7 +404,7 @@ const routeHelpers = require('can-stache-route-helpers');
 Running this transform:
 
 ```bash
-can-migrate '**/*.js' -t version-4/can-stache/scope.js -a
+can-migrate '**/*.stache' -t version-4/can-stache/scope.js -a
```
chasenlehara: This one applies to both types of files.

matthewp:
@mikemitchel Any plans to update this one?

PR #4889, opened 2019-05-23
Update deps for new patch release

bmomberger-bitovi:
Patches up for:
- can-ajax
- can-connect
- can-fixture-socket
- can-list
- can-map
- can-map-define
- can-query-logic
- can-reflect
- can-stache
- can-stache-bindings
- can-stache-route-helpers

Plus major version bumps to the devDependencies that could be upgraded: feathers, semver, github, socket.io-client, steal-conditional


.travis.yml
```diff
@@ -1,5 +1,6 @@
 sudo: false
 language: node_js
+dist: trusty
```
phillipskevin: Did you change this because of the `xvfb` issue? If so, you can fix tit like this: https://github.com/canjs/can-fixture-socket/pull/76/commits/4eca4ee5446e0bab4893c71895fb5270fa51dfe4.

phillipskevin:
The `can-*` packages were updated in https://github.com/canjs/canjs/releases/tag/v5.26.0.

PR #4891, opened 2019-05-23
Adds new packages for minor release

matthewp:
This PR introduces `can-define-object` and `can-define-mixin`. To make
the global.js and ecosystem.js builds work with ES5 I changed the build
process slightly.


bmomberger-bitovi:
Can you also add the dependencies from #4889 into this release?  They're all patches/fixes to the best of my knowledge.

PR #4916, opened 2019-05-24
bumping versions for can-stache-define-element

phillipskevin:


phillipskevin:
Found an issue, so this doesn't need to be release at the moment. Will put in a new PR next week.

PR #4987, opened 2019-05-29
Add can-define-array

matthewp:
This adds can-define-array as an ecosystem package, and updates
associated dependencies.


phillipskevin:
Handling this in https://github.com/canjs/canjs/pull/4996.

PR #4989, opened 2019-05-29
Refine the homepage

jamiemccue:
- [x] Revert font-weight
- [x] Add the social embeds into the “We have your back” section
- [x] Fix CanJS logo jumping down
- [x] Instead of having the diagrams go to the right of the text at wider screen widths, let’s remove this code and force a max-width that keeps the page looking good.


jamiemccue:
@justinbmeyer Can you review the flicker issue?

I hid the class `on-this-page-table` from displaying and that fixes the white line loading issue, however, the CanJS image moves a bit while the classes load. It looks better than before but still not perfect.  

.on-this-page-table {  // fixes the flicker issue on page load.
  display: none;
}

If you're satisfied then you can merge the PR.

justinbmeyer:
Could you make a gif? That way I can review here and not have to pull and build.

jamiemccue:
Gif of loading:

loader

PR #5021, opened 2019-06-06
updating dependencies

phillipskevin:


matthewp:
There's a problem with can-define-object in the production build: https://github.com/canjs/canjs/pull/5019 . You might want to exclude it from your PR for now.

PR #5047, opened 2019-06-12
Fix can-map-define demo

cherifGsoul:
Bring back `can-map-define` demo

Fixes https://github.com/canjs/can-map-define/issues/100

### Before
Screen Shot 2019-06-12 at 8 03 15 PM

### After fixing
Screen Shot 2019-06-12 at 8 02 37 PM

### DIFF
--- a/demos/can-map-define/make-model-year.html
+++ b/demos/can-map-define/make-model-year.html
@@ -1,0 +1,325 @@
+<div id="out"></div>
+<script type="text/stach" id="demo-html">
+	{{# if(this.makes.isResolved) }}
+		<select value:bind="string-to-any(~makeId)">
+			{{^makeId}}
+				<option value='undefined'>Select a Make</option>
+			{{/makeId}}
+			{{# for (carMake of makes.value)}}
+				<option value="{{carMake.id}}">{{carMake.name}}</option>
+			{{/ for }}
+		</select>
+	{{else}}
+		<select disabled><option>Loading...</option></select>
+	{{/if}}
+
+	{{# if(this.modelsPromise) }}
+		{{#if(models)}}
+			<select value:bind="string-to-any(~modelId)">
+				{{^modelId}}
+					<option value='undefined'>Select a Model</option>
+				{{/modelId}}
+				{{# for (carModel of models)}}
+					<option value="{{carModel.id}}">{{carModel.name}}</option>
+				{{/for}}
+			</select>
+		{{else}}
+			<select disabled><option>Loading Models</option></select>
+		{{/if}}
+	{{else}}
+		<select disabled><option>Models</option></select>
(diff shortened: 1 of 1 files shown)


demos/can-map-define/make-model-year.html
```diff
@@ -0,0 +1,325 @@
+<div id="out"></div>
+<script type="text/stach" id="demo-html">
+	{{# if(this.makes.isResolved) }}
+		<select value:bind="string-to-any(~makeId)">
+			{{^makeId}}
+				<option value='undefined'>Select a Make</option>
```
chasenlehara: `value="undefined"` (here and other places)

demos/can-map-define/make-model-year.html
```diff
@@ -0,0 +1,325 @@
...
+	<div>
+		{{# for (vehicle of this.vehicles.value) }}
+			<h2>{{vehicle.name}}</h2>
+			<img src="{{vehicle.thumb}}" width="200px"/>
+		{{/for }}
+	</div>
+</script>
+<script src="../../node_modules/steal/steal.js" dev-bundle main='@empty'>
+	import {DefineMap, DefineList, CanMap, QueryLogic, restModel, fixture, stache, route, stacheConverters } from "can";
+	import "can-map-define";
```
chasenlehara: The indentation here looks off.

demos/can-map-define/make-model-year.html
```diff
@@ -0,0 +1,325 @@
...
+    fixture.delay = 1200;
+
+    var eqeq = function(a, b) {
```
chasenlehara: You can use const/let everywhere.

demos/can-map-define/make-model-year.html
```diff
@@ -0,0 +1,325 @@
...
+                year: 2013,
+                name: "2013 Leaf",
+                thumb: "http://www.blogcdn.com/www.autoblog.com/media/2012/04/01-2013-nissan-altima-ny.jpg"
+            },
+            {
+                id: 8,
+                modelId: 4,
+                year: 2014,
+                name: "2014 Leaf",
+                thumb: "http://images.thecarconnection.com/med/2013-nissan-leaf_100414473_m.jpg"
```
chasenlehara: Change as many of these to https as you can.

demos/can-map-define/make-model-year.html
```diff
@@ -0,0 +1,325 @@
...
+        }
+    });
+
+    var mmy = new MMYViewModel();
+
+    route.data = mmy;
+
+    route.start();
+
+    out.appendChild(stache.from("demo-html")(mmy));
```
chasenlehara: Do we still have to do this, or can we just use a component?

PR #5065, opened 2019-06-20
Add can-dom-mutate tests

### LINKED ISSUE #5024: Add tests for can-dom-mutate to full canjs test suite
The tests for can-dom-mutate should be added to the full test battery for canjs.  However, there are test failures in production and in IE11 when trying to add the tests to the full suite.  These tests need to be corrected before full integration.

Some example failures can be found at https://travis-ci.org/canjs/canjs/builds/542491847

cherifGsoul:
This adds `can-dom-mutate` to full canjs test suite  battery.

Update: Seems the issue was fixed in `can-dom-mutate` v1.3.9

Fixes #5024


package.json
```diff
@@ -80,7 +80,7 @@
     "can-dom-data": "1.0.2",
     "can-dom-data-state": "1.1.1",
     "can-dom-events": "1.3.11",
-    "can-dom-mutate": "1.3.7",
+    "can-dom-mutate": "2.0.2",
```
chasenlehara: A major version bump would mean a major version bump for CanJS, right? I don’t think we should change that yet.

cherifGsoul: Fixed!

PR #5080, opened 2019-06-25
update session guide to use stache-define-element

nlundquist:
Updating the session guide to use the new StacheDefineElement module. 

Adding reviewers to confirm I'm using StacheDefineElement as effectively as possible.


docs/can-guides/topics/data/managing-sessions/app-session-beforeSend.html
```diff
@@ -125,16 +125,18 @@ <h3>Todos:</h3>
         </li>
       {{/each}}
       </ol>
-    `,
-    ViewModel: {
+    `;
+
+    static define = {
```
phillipskevin: `todos` should be using `async(resolve) { ... }`

Issue #5095, opened 2019-07-01
Make sure all fonts have a fallback


chasenlehara:
If the user doesn’t have the right font installed/enabled, the SVG on [this page](https://canjs.com/doc/guides/technology-overview.html#Overview) looks horrible:
<img width="430" alt="Screen Shot 2019-07-01 at 10 27 44 AM" src="https://user-images.githubusercontent.com/10070176/60455344-f9604800-9bea-11e9-90e1-a9ff896f9a4a.png">

For this issue, make sure that any SVGs in the `docs` folder that need a font have an appropriate list of fallback fonts.

ryangwilson:
[SVG OPT.zip](https://github.com/canjs/canjs/files/3347218/SVG.OPT.zip)
Here are all of the SVGs I created last year with their fonts outlined, so this should never be a problem ever ever again. 

I also ran the SVGs through an optimizer.

ryangwilson:
PNGs are closer to the original SVG file sizes. I recommend swapping the SVGs for PNGs (in this case)

[tinified.zip](https://github.com/canjs/canjs/files/3350626/tinified.zip)

ryangwilson:
Scrap all of my previous comments. Chasen and I have discovered a new SVG-as-image creation process that embeds a Google Font and sets a fallback. BONUS - the file size ends up being HALF of what it was before using this new method
<img width="538" alt="Screen Shot 2019-07-03 at 8 19 34 AM" src="https://user-images.githubusercontent.com/4571457/60591437-7961eb80-9d6c-11e9-9aa2-70390139e610.png">

Safari (the please where this was an issue)
<img width="1174" alt="Screen Shot 2019-07-03 at 8 21 05 AM" src="https://user-images.githubusercontent.com/4571457/60591488-91396f80-9d6c-11e9-910a-5fd8a651e15e.png">

With Google Fonts DISABLED
<img width="1174" alt="Screen Shot 2019-07-03 at 8 29 18 AM" src="https://user-images.githubusercontent.com/4571457/60591547-b332f200-9d6c-11e9-80c2-41cb321ed30e.png">

ryangwilson:
Closed by https://github.com/canjs/canjs/pull/5100

PR #5100, opened 2019-07-03
Updates SVG and how SVG is called into the page

ryangwilson:
Fonts now working correctly in Safari
Screen Shot 2019-07-03 at 8 21 05 AM

With a fallback
Screen Shot 2019-07-03 at 8 29 18 AM

BONUS - updated file is 50% smaller than previous
Screen Shot 2019-07-03 at 8 19 34 AM


ryangwilson:
closes #5095

PR #5123, opened 2019-07-10
Update technical.md

Mattchewone:
Fix undeclared variable and missing comma from being valid js syntax

### DIFF
--- a/docs/can-guides/introduction/technical.md
+++ b/docs/can-guides/introduction/technical.md
@@ -180,7 +180,7 @@ let fullName = value.returnedBy(function(){
 or like:
 
 ```js
-Person = DefineMap.extend({
+let Person = DefineMap.extend({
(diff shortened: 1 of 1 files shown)


docs/can-guides/introduction/technical.md
```diff
@@ -180,7 +180,7 @@ let fullName = value.returnedBy(function(){
 or like:
 
 ```js
-Person = DefineMap.extend({
+let Person = DefineMap.extend({
```
chasenlehara: `let` makes me think this is going to be reassigned somewhere. In the future, I think we should try to consistently use `const` unless something’s going to be reassigned.

Mattchewone: Sure, I was going to use `const` but changed my mind after seeing more uses of `let` :( (not a good reason)

PR #5130, opened 2019-07-11
fix link in configuring-requests

Mattchewone:
Fix missing `)` in link to google doc.


Mattchewone:
> Actually, we should take these out, I don’t think they’re needed anymore.

Ok cool, even better. Should we remove the entire "Note to reader" section @chasenlehara ?

chasenlehara:
Yeah I think so. Looks like they’re on other pages too.

PR #5143, opened 2019-07-15
update todomvc-with-steal.md

Mattchewone:
Add comment block around `...`s and add variable declarations for `DefineMap` `DefineList` examples.


docs/can-guides/commitment/recipes/todomvc-with-steal/todomvc-with-steal.md
```diff
@@ -231,23 +231,23 @@ QUnit.equal(todo.complete, true, "toggleComplete works");
 - The [can-define.types.type type] behavior defines a property’s type like:
 
   ```js
-  DefineMap.extend({
+  const Map = DefineMap.extend({
```
chasenlehara: Generally I would avoid `Map` because it’s a built-in constructor.

Mattchewone: Ah good point. Will rename

PR #5157, opened 2019-07-16
Add basic style guide page and link in sidebar

jamiemccue:
- [x] Get a basic working prototype that documents a couple of styles
- [x] Add a style guide page to the [Contributing docs](https://canjs.com/doc/guides/contribute.html)

### DIFF
--- a/docs/can-guides/contribute/site-style-guide.md
+++ b/docs/can-guides/contribute/site-style-guide.md
@@ -1,0 +1,87 @@
+@page guides/contributing/site-style-guide Site Style Guide
+@parent guides/contribute 7
+
+@description
+This is the CanJS Site Style Guide
+
+@body
+
+<h2>Headings</h2>
+<h2>H2</h2>
+<p>Paragraph</p>
+<h2>H2 with <code>code</code></h2>
+<h3>H3</h3>
+<p>Paragraph</p>
+<h4>H4</h4>
+<p>Paragraph</p>
+<h5>H5</h5>
+<p>Paragraph</p>
+<h6>H6</h6>
+<p>Paragraph</p>
+<h2>Block elements</h2>
+<p>A paragraph. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>
+<p>Another paragraph. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
+<ul>
+<li>List element</li>
+<li>Another item</li>
+<li>Another item</li>
+</ul>
+<p>Another paragraph. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
+<ul>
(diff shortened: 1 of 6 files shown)


docs/can-guides/contribute/site-style-guide.md
```diff
@@ -0,0 +1,87 @@
+@page guides/contributing/site-style-guide Site Style Guide
+@parent guides/contribute 7
+
+@description
+This is the CanJS Site Style Guide
```
chasenlehara: This is mostly just repeating the title. Most of the other pages have something like “Learn about X.” I like to think of these descriptions in the context of when someone searches… if someone searches for “style” or “guide” and this is one of the results, will they understand what the page is? I also like to put in other “keywords” that someone might search for that should bring up this page.

docs/can-guides/contribute/site-style-guide.md
```diff
@@ -0,0 +1,87 @@
...
+
+@body
+
+<h2>Headings</h2>
+<h2>H2</h2>
+<p>Paragraph</p>
```
chasenlehara: What’s the benefit of writing out the html vs. writing it in Markdown?

docs/can-guides/contribute/site-style-guide.md
```diff
@@ -0,0 +1,87 @@
...
+<p>HTML code example:</p>
+<pre><code class="language-html">&lt;hello-world&gt;&lt;/hello-world&gt;
+</code></pre>
+<p>JS code example:</p>
+<pre><code class="language-js">console.log(&quot;Hello There&quot;)
+</code></pre>
+<p>CodePen example:</p>
+<pre><code class="language-js">console.log(&quot;Hello There&quot;)
+</code></pre>
+<div class='codepen'></div>
```
chasenlehara: For the CodePen integration, I think it’d be better to have a real example in Markdown because this structure could change (will change?). Otherwise, we’d have to update the style guide _and_ make changes in the bit-docs package.

docs/can-guides/contribute/site-style-guide.md
```diff
@@ -0,0 +1,87 @@
...
+  var message = &quot;Hello There&quot;;
+  console.log( message );
+}
+myFunction();
+function  myFunction(){
+  var message = &quot;Hello There&quot;;
+  console.log( message );
+}
+myFunction();
+</code></pre>
```
chasenlehara: I would replace this with a real example, e.g. `@sourceref ../experiment/crud-beginner/1.js`

I would also add [bash](https://github.com/canjs/canjs/search?q=bash) and [handlebars](https://github.com/canjs/canjs/search?q=handlebars) examples.

docs/can-guides/contribute/site-style-guide.md
jamiemccue: @chasenlehara Can you provide a CodePen example in Markdown?

docs/can-guides/contribute/site-style-guide.md
jamiemccue: @chasenlehara can you provide examples? I'm unaware of how you'd like them formatted.

docs/can-guides/contribute/site-style-guide.md
chasenlehara: We use the `@codepen` tag to make an example able to be opened in CodePen.

An example from: https://github.com/canjs/canjs/blob/91bce285dd8d471541f7d92dedada200ef94b778/docs/can-guides/topics/data/managing-sessions/managing-sessions.md#configuring-the-behavior

```md
@sourceref ./cookie-session-basic.html
@highlight 22-27,only
@codepen
```

docs/can-guides/contribute/site-style-guide.md
chasenlehara: Those links have some examples, and here are some more:

    ```bash
    cd project
    npm install can --save
    ```

    ```handlebars
    {{# for(value of values) }}
        <p>{{ value }}</p>
    {{ else }}
        <p>No items</p>
    {{/ for }}
    ```

Issue #5115, opened 2019-07-09
Fix the counter example in the HTML guide


chasenlehara:
The first time the [HTML guide](https://canjs.com/doc/guides/html.html#Overview) loads, the counter element doesn’t render. It does appear if you refresh the page.

![html page](https://user-images.githubusercontent.com/10070176/60911119-55ece400-a237-11e9-88cf-29313f7f3c5f.gif)

cherifGsoul:
The example is loaded for me in Chrome, Firefox and Safari, tested in private browsing mode also.

chasenlehara:
To be clear, it loads correctly when you 1) open a new tab, 2) go to canjs.com > Guides, and 3) click on the HTML page? Could you please attach a gif like mine that shows this?

PR #5165, opened 2019-07-17
add id attr to counter demo script

cherifGsoul:
Add `id` attribute to the [counter demo](https://canjs.com/doc/guides/html.html) work with the fix in this [PR](https://github.com/canjs/bit-docs-html-canjs/pull/563)

For #5115

### DIFF
--- a/docs/can-guides/experiment/technology/html.md
+++ b/docs/can-guides/experiment/technology/html.md
@@ -31,7 +31,7 @@ Let's say you want to create a page that counts clicks like the following: (_cli
 <style>
   my-counter button {margin-left: 15px;}
 </style>
-<script type="text/steal-module">
+<script id="counter-demo" type="text/steal-module">
 const Component = require("can-component");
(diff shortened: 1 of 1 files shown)


docs/can-guides/experiment/technology/html.md
```diff
@@ -31,7 +31,7 @@ Let's say you want to create a page that counts clicks like the following: (_cli
 <style>
   my-counter button {margin-left: 15px;}
 </style>
-<script type="text/steal-module">
+<script id="counter-demo" type="text/steal-module">
```
chasenlehara: I don’t think we should have to add these IDs for the demos to work correctly. Can this be done dynamically?

cherifGsoul: @chasenlehara it is fixed in https://github.com/canjs/bit-docs-html-canjs/pull/563 I'm closing this PR

PR #5169, opened 2019-07-18
Update the CRUD guide with the renamed APIs

chasenlehara:
The CRUD guide now shows:

- can-observable-array
- can-observable-object
- can-stache-element (with props)


chasenlehara:
Oh no worries, this was in `major` already so I wasn’t expecting it to be updated!

justinbmeyer:
I'd _really_ like to avoid the getters.  I'm happy to make codepen use babel.

phillipskevin:
@Mattchewone and I tried this and it doesn't work. If you enable the babel preprocessor in codepen it tries to transpile the import:

image

Another option is showing `static view = ``;` but having the codepen plugin transform it into a getter. This would take some work though.

PR #5177, opened 2019-07-29
Add RecursiveObservable

matthewp:


justinbmeyer:
Did we finalize what these things were going to be?

matthewp:
RecursiveObservable we have (unless there was an objection I wasn't aware of), the array shorthand is discussed here: https://github.com/canjs/can-observable-array/issues/35#issuecomment-516092361 and not finalized.

matthewp:
Sounds like DeepObservable was decided on

PR #5183, opened 2019-08-01
Add docs for the .bit-docs-screenshot class to the style guide

chasenlehara:
Screen Shot 2019-08-01 at 11 40 37 AM

### DIFF
--- a/docs/can-guides/contribute/site-style-guide.md
+++ b/docs/can-guides/contribute/site-style-guide.md
@@ -121,3 +121,15 @@ npm install can --save
 
 The following is a very long sentence that will hopefully go across many lines because it
 is so long and filled with <code>code elements</code>, <strong>bold elements</strong>, <em>italic elements</em>, <a href="#">link elements</a>.
+
+## Screenshots
+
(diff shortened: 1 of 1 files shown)


docs/can-guides/contribute/site-style-guide.md
```diff
@@ -121,3 +121,15 @@ npm install can --save
...
 The following is a very long sentence that will hopefully go across many lines because it
 is so long and filled with <code>code elements</code>, <strong>bold elements</strong>, <em>italic elements</em>, <a href="#">link elements</a>.
+
+## Screenshots
+
+Use the `bit-docs-screenshot` class on images to center them. Add a `width` to the image to set a max width.
+
+### With width
+
+<img alt="" class="bit-docs-screenshot" src="https://canjs.com/docs/can-guides/images/devtools/panel-component-selected.png" width="600px" />
```
justinbmeyer: wouldn't you set `max-width`?

chasenlehara: No, I don’t think so (and I don’t think that attribute exists in html). Ideally we would have height and width set for all images. The CSS sets the `max-width` to 100% so it’s dynamic.

Issue #5119, opened 2019-07-09
Upgrade the docs for CanJS 6


chasenlehara:
Work should be done against this branch: https://github.com/canjs/canjs/tree/major

- [x] https://canjs.com/doc/guides/crud-beginner.html
- [ ] https://canjs.com/doc/guides/technical.html
- [ ] https://canjs.com/doc/guides/setup.html
- [ ] https://canjs.com/doc/guides/technology-overview.html
- [ ] https://canjs.com/doc/guides/html.html
- [ ] https://canjs.com/doc/guides/routing.html
- [ ] https://canjs.com/doc/guides/data-introduction.html
- [ ] https://canjs.com/doc/guides/data-configuring-requests.html
- [ ] https://canjs.com/doc/guides/data-managing-sessions.html
- [ ] https://canjs.com/doc/guides/data-customizing-connections.html
- [ ] https://canjs.com/doc/guides/debugging.html
- [ ] https://canjs.com/doc/guides/forms.html
- [ ] https://canjs.com/doc/guides/testing.html
- [ ] https://canjs.com/doc/guides/logic.html
- [ ] https://canjs.com/doc/guides/chat.html
- [ ] https://canjs.com/doc/guides/todomvc.html
- [ ] https://canjs.com/doc/guides/recipes/todomvc-with-steal.html
- [ ] https://canjs.com/doc/guides/recipes/canvas-clock.html
- [ ] https://canjs.com/doc/guides/recipes/credit-card-simple.html
- [ ] https://canjs.com/doc/guides/recipes/file-navigator-simple.html
- [ ] https://canjs.com/doc/guides/recipes/signup-simple.html
- [ ] https://canjs.com/doc/guides/recipes/video-player.html
- [ ] https://canjs.com/doc/guides/recipes/cta-bus-map.html
- [ ] https://canjs.com/doc/guides/recipes/modals.html
- [ ] https://canjs.com/doc/guides/recipes/text-editor.html
- [ ] https://canjs.com/doc/guides/recipes/tinder-carousel.html
- [ ] https://canjs.com/doc/guides/recipes/credit-card-advanced.html
- [ ] https://canjs.com/doc/guides/recipes/file-navigator-advanced.html
- [ ] https://canjs.com/doc/guides/recipes/playlist-editor.html
- [ ] https://canjs.com/doc/guides/recipes/search-list-details.html
- [ ] https://canjs.com/doc/guides/api.html
- [ ] https://canjs.com/doc/api.html

[...]

PR #5203, opened 2019-08-14
Remove access to `viewModel` property on the element from docs

### LINKED ISSUE #5138: Update docs for access `viewModel` on the element
We currently show accessing `viewModel` property on the element, with the new `StacheDefineElement` the element is the observable so there is no need to use the `viewModel` property.

```js
const appVM = window.appVM = document.querySelector("todo-mvc").viewModel;
```

Would be:
```js
const appVM = window.appVM = document.querySelector("todo-mvc");
```

We use these in the following guides:

- https://canjs.com/doc/guides/recipes/cta-bus-map.html#Changetheapptitle
- https://canjs.com/doc/guides/recipes/tinder-carousel.html#Showthecurrentandnextprofileimages
- https://canjs.com/doc/guides/todomvc.html#Defineandusethemaincomponent
- https://canjs.com/doc/guides/routing.html#Overview

m-mujica:
Closes #5138

### DIFF
--- a/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
+++ b/docs/can-guides/commitment/recipes/cta-bus-map/cta-bus-map.md
@@ -218,7 +218,7 @@ In this section, we will:
 Run the following in the `Console` tab:
 
 ```js
-document.querySelector('my-component').viewModel.someValue = "TITLE UPDATED";
+document.querySelector('my-component').someValue = "TITLE UPDATED";
 ```
 
 You should see the title update.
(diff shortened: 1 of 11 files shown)


m-mujica:
should I also update stuff like:

```js
events: {
  click() {
      this.viewModel.visible = !this.viewModel.visible;
   }
}
```

or will the codemods take care of `this.viewModel` references? @Mattchewone @phillipskevin

matthewp:
These need to be converted to use `StacheElement` as well.

m-mujica:
hmm got it, should I just use the codemod? I thought that was part of a different issue?

matthewp:
You can use the codemod or do it manually, but I don't think we should merge changes that result in the demo being broken (even if being fixed later).

m-mujica:
roger that

Mattchewone:
Pretty sure the codemods won't change those references.

phillipskevin:
We can do the codemods as part of https://github.com/canjs/canjs/issues/5119.

PR #5215, opened 2019-08-15
update debugging docs

Mattchewone:
Update references to `ObservableObject` from `DefineMap`.

### DIFF
--- a/docs/can-guides/topics/debugging.md
+++ b/docs/can-guides/topics/debugging.md
@@ -158,7 +158,7 @@ If you are using [can-observable-object], name your class:
 import { ObservableObject } from "can";
 
 export default class TheNameOfMyType extends ObservableObject {
-  static props = { /* ... */ }
+  static props = { /* ... */ };
 }
 ```
 
@@ -177,10 +177,10 @@ expressions:
 
 ```js
 // INSTEAD OF THIS:
-map.on("key", function(ev, newVal) { /* ... */ })
+map.on("key", function(ev, newVal) { /* ... */ });
 
 // DO THIS:
-map.on("key", function keyChanged(ev, newVal) { /* ... */ })
+map.on("key", function keyChanged(ev, newVal) { /* ... */ });
 ```
 
 Similarly, if you create [can-compute]s or [can-observation]s yourself, make sure the function
@@ -205,7 +205,7 @@ new Observation(function fullName(){
 >   static props = {
 >     fullName: {
 >       get () { return this.first + " " + this.last; }	
->     }
+>     };
 >   }
 > }
 > ```
@@ -231,7 +231,7 @@ const info = new Observation(function updateInfo(){
 
 info.on(function onInfoChanged(newVal){
     debugger;
-})
+});
 
 person.age = 22;
 ```
@@ -254,7 +254,7 @@ task is enqueued and flushed. Often, you only want to log when
 tasks are run. This can be done with:
 
 ```js
-can.queues.log("flush")
+can.queues.log("flush");
 ```
 
 Both `queues.logStack()` and `queues.log()` log the function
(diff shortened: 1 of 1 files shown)


docs/can-guides/topics/debugging.md
```diff
@@ -197,14 +198,16 @@ new Observation(function fullName(){
 });
 ```
 
-> **Note:** If your function is a property on an observable map or list like [can-define/map/map],
+> **Note:** If your function is a property on an observable class like [can-observable-object],
 > you don't have to name it. For example, CanJS will name the `fullName` getter in the following example:
```
phillipskevin: Did you double check that this is true?

docs/can-guides/topics/debugging.md
```diff
@@ -141,24 +141,25 @@ CanJS tries to create useful names to help identify the objects and
...
-name your function by passing `.extend` a name as the first argument:
+If you are using [can-observable-object], name your class:
 
 ```js
-import { DefineMap } from "can";
+import { ObservableObject } from "can";
 
-export default DefineMap.extend("TheNameOfMyType", { /* ... */ })
+export default class TheNameOfMyType extends ObservableObject {
+  static props = { /* ... */ }
```
phillipskevin: can you add a semicolon `static props = { /* ... */ };`

docs/can-guides/topics/debugging.md
```diff
@@ -197,14 +198,16 @@ new Observation(function fullName(){
...
 > ```js
-> const Person = DefineMap.extend("Person",{
->   fullName: {
->     get: function(){ return this.first + " " + this.last; }
+> class TheNameOfMyType extends ObservableObject {
+>   static props = {
+>     fullName: {
+>       get () { return this.first + " " + this.last; }	
+>     }
 >   }
```
phillipskevin: semicolon

docs/can-guides/topics/debugging.md
```diff
@@ -345,29 +348,32 @@ Stache templates also have access the [can-stache.helpers.console] methods, maki
...
+	static props = {
+		count: {
+			value({resolve}) {
+				let count = resolve(0);
+				setInterval(() => {
+					resolve(++count);
+				}, 1000);
+			}
+		}
+	}
```
phillipskevin: semicolon.

docs/can-guides/topics/debugging.md
```diff
@@ -345,29 +348,32 @@ Stache templates also have access the [can-stache.helpers.console] methods, maki
...
-                    resolve(++count);
-                }, 1000);
-            }
-        }
-    }
-});
+class MyCounter extends StacheElement {
+	static view = `
+		{{ console.log(count) }}
+	`
```
phillipskevin: semicolon.

docs/can-guides/topics/debugging.md
Mattchewone: Using this snippet:
```js
class MyEl extends StacheElement {
  static view = `
  {{ this.myProp }}
  <input value:bind="this.prop">
  `;
  
  static props = {
    prop: {
      type: type.convert(String),
      bind: fromAttribute
    },
    myProp: {
      get: function () {
        return this.prop
      }
    }
  };
}
customElements.define('my-el', MyEl);
```

When setting a breakpoint I see this output:
```
NOTIFY ran task: Observation<MyEl{}'s myProp getter>.onDependencyChange {fn: ƒ, context: my-el, args: Array(2), meta: {…}}
can-log.js:98 DERIVE ran task: Observation<MyEl{}'s myProp getter>.update
```
I'm assuming this is the expected right?

phillipskevin: Interesting. I assume this is because of the proxy? We should investigate.

phillipskevin:
Will review after publishing.

PR #5216, opened 2019-08-16
Merge master to major

chasenlehara:


pre-document.sh
```diff
@@ -0,0 +1 @@
+#!/bin/bash
```
chasenlehara: Instead of removing this file because it’s not needed in major, I left it so we can reuse it again if we need to for CanJS 7.

PR #5222, opened 2019-08-16
Apply some code style changes to the CTA Bus recipe

m-mujica:


phillipskevin:
Will review after publishing.

Issue #5207, opened 2019-08-14
Remove can-define usage in CanJS core


phillipskevin:
Although, `can-define` was removed from `core.js`, the core build still included `can-define` because it is a dependency of a few other `core` packages:

- [can-rest-model](https://github.com/canjs/can-rest-model/blob/90917099932d1b394081beebf7fd4bbc89138ab1/can-rest-model.js#L16-L21)
- [can-realtime-rest-model](https://github.com/canjs/can-realtime-rest-model/blob/8f05f16e2d0655f317cd84a9b45b7db32edd5411/can-realtime-rest-model.js#L20-L26)
- [can-route/routedata](https://github.com/canjs/can-route/blob/dadda0b603edbbd8177e99d8eaab0a08b9295d67/src/routedata.js#L4)

We should consider changing these to use `ObservableObject` and `ObservableArray`.

## Benefits

1. Documentation will be more consistent

Things like this (from the homepage) mention `Map`, where everything else is using `Object`/`Array`:

![image](https://user-images.githubusercontent.com/5851984/63048347-fa4bf100-be9b-11e9-8d9d-25923d0fabeb.png)

2. Including `can-define` in core makes the build larger

## Open questions

1. Would this break IE11 support for people using `can-component` and `can-define`?
  a. Should we run more of the test suite in IE11? (currently we only run the steal development tests)
2. Would this break users' apps in unexpected ways? (missing methods like `get`,`set`, etc)

matthewp:
It's also in can-component. I'd probably change can-component first and run against IE11 and see what happens. I think that's probably the biggest risk.

phillipskevin:
But `can-component` is no longer in core in 6.0. We want people to be able to continue using `can-component`/`can-define` if they need to support IE11.

phillipskevin:
We are going to do this for 6.0. For the open questions:

1. We will provide `-legacy` packages that are needed for people to keep using `can-component` and `can-define` based apps in IE11
2. These are unlikely as people will only be using `ObservableObject` instead of `DefineMap` if they don't provide their own `Map` option and don't use the `-legacy` packages. In these cases, they can use one of these work-arounds. We can also add these methods to `ObservableObject` later if we decide to do that.

chasenlehara:
[...]

PR #5252, opened 2019-09-03
Add the can-define-realtime-rest-model and can-realtime-rest-model packages

chasenlehara:
Both are listed as legacy in the docs:

Screen Shot 2019-09-03 at 9 35 28 PM

Part of https://github.com/canjs/canjs/issues/5207


justinbmeyer:
Why legacy?

justinbmeyer:
oh .... `define-`

PR #5254, opened 2019-09-04
update session guide to use ObservableObject

nlundquist:
All copy updates complete. However requires minor changes in a dependency for example code to work. 

canjs/can-define-connected-singleton needs to be updated to work with ObservableObject. Might as well rename the project to can-connected-singleton too.

Should only need an additional clause in the if statement [here](https://github.com/canjs/can-define-connected-singleton/blob/master/can-define-connected-singleton.js#L190) so when `Obj` is an instance of `ObservableObject` that `makeSingleton` runs immediately rather than returning a wrapper function.

These codepens will likely also require a version of CanJS that includes the changes [here]( https://github.com/canjs/can-observable-mixin/pull/136) since `dispatch` is called on the `ObservableObject` type.

### DIFF
--- a/docs/can-guides/topics/data/managing-sessions/app-session-beforeSend.html
+++ b/docs/can-guides/topics/data/managing-sessions/app-session-beforeSend.html
@@ -5,15 +5,18 @@
     connect,
     connectCanSession,
     fixture,
+    ObservableObject,
     restModel,
-    type,
-    DefineMap,
     StacheElement,
-  } from "//unpkg.com/can@5/ecosystem.mjs";
+    type,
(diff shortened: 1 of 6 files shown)


justinbmeyer:
Should this depend on ObservableObject? Is everything possible through symbols?

phillipskevin:
@justinbmeyer do you mean the guide would show how to construct your own `Session` and `Todo` observables?

matthewp:
These changes were applied in https://github.com/canjs/canjs/pull/5322 and https://github.com/canjs/canjs/pull/5323

PR #5255, opened 2019-09-04
Upgrade CRUD beginner guide to Can 6

m-mujica:
* Starter pen: https://codepen.io/bitovi/pen/VwZrYyG
* Final pen: https://codepen.io/bitovi/pen/yLBPLgw?editors=0010

* Once this PR is merged https://github.com/canjs/can-demo-models/pull/6, I can do a major release and update the pens

Screen Shot 2019-09-04 at 10 36 13 AM


m-mujica:
@phillipskevin @chasenlehara do you know if the usage of `can@everything` was intentional? so that beginners can import whatever without knowing about different builds? I changed it to `core` but not sure if that's "right".

Issue #3371, opened 2017-06-29
Add file size tests


chasenlehara:
We advertise that CanJS’s core is [under 50KB (gzipped)](https://canjs.com/doc/guides/technical.html#PhenomenalFeatures_SmallSize), but we don’t have any tests to verify that.

andrejewski:
We could look into this if we didn't want to hand-roll something: https://github.com/siddharthkp/bundlesize

PR #5277, opened 2019-09-10
Add tests for the size of core

chasenlehara:
This adds a package called `bundlesize`. Given a list of files, it gzip’s them and checks whether they’re smaller than the sizes configured in `package.json`. It also integrates with GitHub; you can see the integration in this PR!

I’ve added tests for three files:

- `core.min.mjs` (currently 100.83KB gzip’d)
- `core.mjs` (currently 310KB gzip’d)
- `dist/global/core.js` (currently 181.03KB gzip’d)

Maybe we want to test more files or fewer files, but for now I’ve added those because I’ll want automated confirmation that the core size actually decreases when https://github.com/canjs/canjs/issues/5207 is done.

Closes https://github.com/canjs/canjs/issues/3371

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test": "npm run test-development && npm run test-production && npm run test-builders && npm run test-global-build",
     "test-builders": "npm run build-webpack-test && npm run testee-builders",
     "test-development": "npm run testee",
-    "test-global-build": "npm run build && npm run testee-global-build",
+    "test-global-build": "npm run build && bundlesize && npm run testee-global-build",
     "test-local": "npm run build && npm run testee",
     "test-production": "npm run build-tests && npm run testee-production",
     "test-sauce-labs": "node test/test-sauce-labs.js",
@@ -155,6 +155,7 @@
     "@feathersjs/socketio-client": "^1.2.1",
     "@octokit/rest": "^16.27.3",
     "bit-docs": "0.2.0",
+    "bundlesize": "^0.18.0",
     "can-reflect-tests": "^1.0.0",
     "can-test-helpers": "^1.1.1",
     "core-js": "^2.5.7",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "test": "npm run test-development && npm run test-production && npm run test-builders && npm run test-global-build",
     "test-builders": "npm run build-webpack-test && npm run testee-builders",
     "test-development": "npm run testee",
-    "test-global-build": "npm run build && npm run testee-global-build",
+    "test-global-build": "npm run build && bundlesize && npm run testee-global-build",
```
chasenlehara: I tacked it onto this build because everything is already being built for these tests, but maybe there’s a better place / it should be split off into its own test.

package.json
```diff
@@ -296,5 +297,19 @@
...
+      "path": "./core.min.mjs",
+      "maxSize": "101 kB"
+    },
+    {
+      "path": "./core.mjs",
+      "maxSize": "311 kB"
+    },
+    {
+      "path": "./dist/global/core.js",
+      "maxSize": "182 kB"
```
chasenlehara: All of these sizes are within 1 kB of the current sizes. Maybe these should allow for more growth? I personally think that erring on the side of being too restrictive is preferable because we might otherwise not notice these files growing a bit at a time if we have, say, 5 or 10 kB of headroom.

justinbmeyer:
Could we take 2 days to try to reduce this?  I remember there being some easy-ish wins.

package.json
matthewp: If we're going to make file size a priority, which we haven't in the past, then having little to know headroom is a good way to go. Every file size increase should be approved.

matthewp:
Can do this in another PR, but it would also be nice to get the size of a smoke test after the production build. I remember it being smaller than this, probably because of tree shaking.

PR #5280, opened 2019-09-11
Upgrade Tinder recipe to Can 6

m-mujica:
Screen Shot 2019-09-11 at 11 40 18 AM


m-mujica:
@phillipskevin this guide loads https://www.npmjs.com/package/pepjs-improved because mobile safari does not support pointer events.

Seems like they added support last year https://caniuse.com/#feat=pointer-events... Should we remove it from the docs and codepen?

phillipskevin:
If it is as simple as removing the script tags, let's do it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "can",
-  "version": "6.0.0-pre.10",
+  "version": "6.0.0-pre.11",
```
phillipskevin: Did `major` not get pushed?

PR #5291, opened 2019-09-16
Upgrade to can-type 1.0.0-pre.8

matthewp:


matthewp:
This increases the bundlesize for core by 80 bytes. This is for the
`type.all` and `type.convertAll` feature.

PR #5303, opened 2019-09-20
Update can-type and can-stache-element

m-mujica:


m-mujica:
@phillipskevin what's the procedure in this case?

phillipskevin:
For now just update the size in the package.json

PR #5305, opened 2019-09-23
Update the Testing guide for CanJS 6.x

chasenlehara:
This is a big overhaul of the testing guide:

- All the new APIs (can-observable-object, can-stache-element, etc.) replace all the old APIs
- A lot of the language is re-written to revolve around custom elements
- All of the examples are taken out of the main Markdown file and put into individual HTML files that can be more easily reused throughout the document

Part of https://github.com/canjs/canjs/issues/5119


phillipskevin:
I know I merged this already, but I have some feedback:

I think the top heading should just be `Observables` and have a sentence explaining that these techniques work for `ObservableObject` and `ObservableArray`.
image



We should keep the `Custom Elements` section called `Components` and change the `ObservableObject` header to `Properties`:

> All of the techniques described in Testing Observables can be used for testing a StacheElement's properties by creating an instance of the component:

image


In the `DOM Events` section, can you add an explanation for 

> 2. Calling render on the instance

something like

> 2. Calling render on the instance to render its `view` into the element's `innerHTML`

The section on `connect` should say `connected`. User puts their code in `connected`, calls `connect` to test it.

image

PR #5404, opened 2019-10-28
Fix search result is obscured issue

cherifGsoul:
Fixes https://github.com/canjs/bit-docs-html-canjs/issues/583

The result after the fix in the GIF:

search

### DIFF
--- a/docs/can-canjs/canjs.md
+++ b/docs/can-canjs/canjs.md
@@ -32,7 +32,7 @@ h3 {
   color: #282C33;
   margin-bottom: 10px;
 }
-code:before {
+code[class*=language-]:before {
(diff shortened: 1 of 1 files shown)


docs/can-canjs/canjs.md
```diff
@@ -32,7 +32,7 @@ h3 {
   color: #282C33;
   margin-bottom: 10px;
 }
-code:before {
+code[class*=language-]:before {
```
chasenlehara: This fixes the issue, but I think the bigger issue is that all these styles apply to the rest of the page.

I think we should scope all these styles to `.body` so they only apply to the homepage.

PR #5488, opened 2020-05-15
Version and release integration

cherifGsoul:
Add release notes script and the custom template.


scripts/generate-release-notes.js
```diff
@@ -0,0 +1,30 @@
...
+const template = require('./release-template');
+
+const options = {
+	token: "",
+	owner: "canjs",
+	repo: "canjs",
+	template
+  };
+
+const output = getReleaseNotes('v6.3.0', 'v6.4.0', options);
```
phillipskevin, line 26: The versions and token should come from command line arguments.

cherifGsoul: @phillipskevin Done!

PR #5510, opened 2020-10-19
Removed link on "enterprise companies"

hkutcher:
"enterprise companies" pointed to https://canjs.com/doc/guides/who-uses.html, which resulted in a 404 page. In a Slack conversation, Kevin suggested removing the link, since the page it points to is pretty outdated.


cherifGsoul:
Thank you @hkutcher

